PR #1, opened 2017-10-10
Add cfengine-lynis to package index

nickanderson:


olehermanse:
Cool, let's try this!

PR #8, opened 2021-07-14
Added function `pretty` for pretty printing JSON

larsewi:
Signed-off-by: larsewi <lars.erik.wik@northern.tech>

### DIFF
--- a/cfbs/pretty.py
+++ b/cfbs/pretty.py
@@ -1,0 +1,97 @@
+import json
+from collections import OrderedDict
+
+
+def pretty_string(s):
+    s = json.loads(s, object_pairs_hook=OrderedDict)
+    return pretty(s)
+
+
+def pretty(o):
+    MAX_LEN = 80
+    INDENT_SIZE = 2
+
+    def _should_wrap(parent):
+        assert isinstance(parent, (tuple, list, dict))
+
+        if isinstance(parent, dict):
+            parent = parent.values()
+
+        count = 0
+        for child in parent:
+            if isinstance(child, (tuple, list, dict)):
+                if len(child) >= 2:
+                    count += 1
+        return count >= 2
+
+    def _encode_list(lst, indent, cursor):
+        if not lst:
+            return "[]"
+
+        if not _should_wrap(lst):
+            buf = json.dumps(lst)
+            assert "\n" not in buf
+            if indent + cursor + len(buf) <= MAX_LEN:
+                return buf
+
+        indent += INDENT_SIZE
+        buf = "[\n" + " " * indent
+        first = True
+        for value in lst:
+            if first:
+                first = False
+            else:
+                buf += ",\n" + " " * indent
+            buf += _encode(value, indent, 0)
+        indent -= INDENT_SIZE
+        buf += "\n" + " " * indent + "]"
+
+        return buf
+
+    def _encode_dict(dct, indent, cursor):
+        if not dct:
+            return "{}"
+
+        if not _should_wrap(dct):
+            buf = json.dumps(dct)
+            buf = "{ " + buf[1 : len(buf) - 1] + " }"
+            assert "\n" not in buf
+            if indent + cursor + len(buf) <= MAX_LEN:
+                return buf
+
+        indent += INDENT_SIZE
+        buf = "{\n" + " " * indent
+        first = True
+        for key, value in dct.items():
+            if first:
+                first = False
+            else:
+                buf += ",\n" + " " * indent
+            if not isinstance(key, str):
+                raise ValueError("Illegal key type '" + type(key).__name__ + "'")
+            entry = '"' + key + '": '
+            buf += entry + _encode(value, indent, len(entry))
+        indent -= INDENT_SIZE
+        buf += "\n" + " " * indent + "}"
+
+        return buf
+
+    def _encode(data, indent, cursor):
+        if data is None:
+            return "null"
+        elif data is True:
+            return "true"
+        elif data is False:
+            return "false"
+        elif isinstance(data, (int, float)):
+            return repr(data)
+        elif isinstance(data, str):
+            return '"' + data + '"'
+        elif isinstance(data, (list, tuple)):
+            return _encode_list(data, indent, cursor)
+        elif isinstance(data, dict):
+            return _encode_dict(data, indent, cursor)
+        else:
+            raise ValueError("Illegal value type '" + type(data).__name__ + "'")
+
+    return _encode(o, 0, 0)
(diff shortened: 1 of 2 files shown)


cf-bottom:
Thanks for submitting a PR! Maybe @vpodzime can review this?

cfbs/pretty_json.py
```diff
@@ -0,0 +1,114 @@
+import json
+from typing import OrderedDict
+
+
+def pretty(s):
+    MAX_LEN = 80
+    INDENT_SIZE = 2
+
+    def _force_wrap(parent):
```
olehermanse, line 9: `should_wrap` or `should_force_wrap` like a question.

cfbs/pretty_json.py
```diff
@@ -0,0 +1,114 @@
...
+        if len(parent) < 2:
+            return False
+        for child in parent:
+            if not isinstance(child, (tuple, list, dict)):
+                return False
+            if len(child) < 2:
+                return False
```
olehermanse, line 16: Count number of children with 2 or more elements. If count is more than 1, wrap.

cfbs/pretty_json.py
```diff
@@ -0,0 +1,114 @@
...
+                    first = False
+                else:
+                    buf += ", "
+                if not isinstance(key, str):
+                    raise ValueError("Illegal key type '" + type(key).__name__ + "'")
+                buf += '"' + key + '": '
+                buf += _encode(value, indent, cursor + len(buf))
+            buf += " }"
+
+            if indent + cursor + len(buf) <= MAX_LEN:
```
olehermanse, line 73:
```suggestion
            assert "\n" not in buf
            if indent + cursor + len(buf) <= MAX_LEN:
```

cfbs/pretty_json.py
```diff
@@ -0,0 +1,114 @@
...
+                    break
+                if first:
+                    first = False
+                else:
+                    buf += ", "
+                if not isinstance(key, str):
+                    raise ValueError("Illegal key type '" + type(key).__name__ + "'")
+                buf += '"' + key + '": '
+                buf += _encode(value, indent, cursor + len(buf))
+            buf += " }"
```
olehermanse, line 71: If this is only for 1 line anyway, could you just use `json.dumps()`?

olehermanse: or map or list / dict comprehension.

cfbs/pretty_json.py
```diff
@@ -0,0 +1,114 @@
...
+                    break
+                if first:
+                    first = False
+                else:
+                    buf += ", "
+                buf += _encode(value, indent, cursor + len(buf))
+            buf += "]"
+
+            if indent + cursor + len(buf) <= MAX_LEN:
+                return buf
```
olehermanse, line 37: Unclear whether buf can be multiple lines or not. Please change to json.dumps, or add asserts and map/join.

olehermanse (changes requested):

cfbs/pretty.py
```diff
@@ -0,0 +1,96 @@
+import json
+from json import encoder
+from typing import OrderedDict
```
vpodzime, line 3: `from collections import OrderedDict`

cfbs/pretty.py
```diff
@@ -0,0 +1,96 @@
...
+        assert isinstance(parent, (tuple, list, dict))
+
+        if isinstance(parent, dict):
+            parent = parent.values()
+
+        count = 0
+        for child in parent:
+            if isinstance(child, (tuple, list, dict)):
+                if len(child) >= 2:
+                    count += 1
```
vpodzime, line 20: I'd prefer a bit more functional-like approach here, something like:
```
count = len([child for child in parent if isinstance(child, (tuple, list, dict) and len(child) >= 2])
return rount >= 2
```
But YMMV.

cfbs/pretty.py
```diff
@@ -0,0 +1,96 @@
...
+        if data is None:
+            return "null"
+        elif data is True:
+            return "true"
+        elif data is False:
+            return "false"
+        elif isinstance(data, int):
+            return int.__repr__(data)
+        elif isinstance(data, float):
+            return float.__repr__(data)
```
vpodzime, line 85: You could merge these two cases into
```
elif isinstance(data, (int, float)):
    return repr(data)
```

test/test_pretty.py
```diff
@@ -0,0 +1,87 @@
...
+            '{ "This": "line", "is": "equal", "to": 80, "characters": "dont\'t wrap me...." }'
+        )
+        == '{ "This": "line", "is": "equal", "to": 80, "characters": "dont\'t wrap me...." }'
+    )
+    assert (
+        pretty(
+            '{ "This": "line", "is": "greater", "than": 80, "characters": "wrap me ........" }'
+        )
+        == '{\n  "This": "line",\n  "is": "greater",\n  "than": 80,\n  "characters": "wrap me ........"\n}'
+    )
```
vpodzime, line 41: I'd prefer this multiline cases to be checked like this:
```
expected = """{
  "This": "line",
  "is": "greater",
  ...
}"""
assert(pretty(...) == expected)
```

vpodzime (commented):
Looks good to me otherwise. Please change the unit test to show the structure of the expected format better. The current format makes it very hard to review.

cfbs/pretty.py
```diff
@@ -0,0 +1,96 @@
...
+        elif isinstance(data, str):
+            return '"' + data + '"'
+        elif isinstance(data, (list, tuple)):
+            return _encode_list(data, indent, cursor)
+        elif isinstance(data, dict):
+            return _encode_dict(data, indent, cursor)
+        else:
+            raise ValueError("Illegal value type '" + type(data).__name__ + "'")
+
+    data = json.loads(s, object_pairs_hook=OrderedDict)
```
olehermanse, line 95: Should be possible to call a pretty function with a dictionary. 9/10 times I want to generate some dictionary, and then prettify it before saving to file or printing to terminal.

Maybe it makes sense to copy what the json module is doing (`load()`, `loads()`, `dump()`, `dumps()`). You could have `pretty()` and `pretty_string()`.

olehermanse (changes requested):

cfbs/pretty.py
larsewi: @vpodzime, @olehermanse  how about ?
```Python
count = len(
    filter(
        lambda child: isinstance(child, (tuple, list, dict))
        and len(child) >= 2,
        parent,
    )
)
```

olehermanse (approved):
LGTM, just a small comment.

PR #10, opened 2021-07-20
Added cfbs pretty command and removed prettier dependency

olehermanse:


cfbs/commands.py
```diff
@@ -78,6 +78,18 @@ def get_index(prefer_offline=False) -> dict:
...
+def pretty_command(filenames: list) -> int:
+    if not filenames:
+        user_error("Filenames missing for cfbs pretty command")
+    for f in filenames:
+        if not f or not f.endswith(".json"):
+            user_error(f"cfbs pretty command can only be used with .json files, not '{os.path.basename(f)}'")
+        try:
+            pretty_file(f)
+        except FileNotFoundError:
+            user_error(f"File '{f}' not found")
```
larsewi, line 90: How about other exceptions, e.g. permission denied? A stack trace is not very user friendly. You could use the `except Exception` to catch all others?

olehermanse: `except Exception` is generally avoided because it can catch too much. (For example a HTTPException that you are trying to catch further up the call stack). The best approach is to add the exceptions you expect explicitly, and give a nice error message. When you encounter a stack trace you add another except clause with a nice error message for that specific case. In many cases, when you haven't added a nice and specific error message, it is better to get a full stacktrace - it's not so helpful to a user, but very nice for the programmer who is going to fix the issue.

I could've added one for permissions, but I didn't bother. I added file not found because it's very likely that people will encounter it. Feel free to submit a PR ;)

PR #26, opened 2021-10-11
CFE-3779 Show STDOUT/STDERR in case of command failure

craigcomstock:
Should help for autoconf not installed for masterfiles module
as well as any other commands which fail.

Ticket: CFE-3779
Changelog: title


cf-bottom (changes requested):
Please use a company e-mail instead of c***@unreasonablefarm.org

PR #27, opened 2021-10-13
CFE-3764 Added show/info command

craigcomstock:
Ticket: CFE-3764
Changelog: title

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -1,4 +1,7 @@
-#!/usr/bin/env python3
+"""
+Functions ending in "_command" are dynamically included in the list of commands
+in main.py for -h/--help/help.
+"""
 import os
 
 from cfbs.utils import (
@@ -31,6 +34,8 @@ def get_definition() -> dict:
     global definition
     if not definition:
         definition = read_json(cfbs_filename())
+    if not definition:
+        user_error("Unable to read {}".format(cfbs_filename()))
     return definition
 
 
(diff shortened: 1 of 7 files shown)


cfbs/commands.py
```diff
@@ -460,5 +464,64 @@ def install_command(destination=None) -> int:
...
+
+def info_command(modules, index=None):
+    if not index:
+        index = get_index_from_config()
+    index = Index(index)
+    build = get_definition()["build"]
+    alias = None
+
+    for module in modules:
+        print()  # whitespace for readability
```
vpodzime, line 480: Should there be whitespace before the first module info?

cfbs/commands.py
```diff
@@ -460,5 +464,64 @@ def install_command(destination=None) -> int:
...
+        if not index.exists(module):
+            print("Module '{}' does not exist".format(module))
+            continue
+        if module in index:
+            data = index[module]
+            if "alias" in data:
+                alias = module
+                module = data["alias"]
+                data = index[module]
+            found = [m for m in build if m["name"] == module]
```
vpodzime, line 490: `found = any(m for m...)`

cfbs/commands.py
```diff
@@ -460,5 +464,64 @@ def install_command(destination=None) -> int:
...
+            data["status"] = "Added" if len(found) > 0 else "Not Added"
```
vpodzime, line 491: `if found`

cfbs/commands.py
```diff
@@ -460,5 +464,64 @@ def install_command(destination=None) -> int:
...
+        else:
+            if not module.startswith("./"):
+                module = "./" + module
+            data = [m for m in build if m["name"] == module]
```
vpodzime, line 495: `data = next((m for m...), None)` and then `if data is None:...`

cfbs/commands.py
```diff
@@ -460,5 +464,64 @@ def install_command(destination=None) -> int:
...
+            "status",
+            "by",
+            "tags",
+            "repo",
+            "commit",
+            "dependencies",
+            "added_by",
+            "description",
+        ]
+        data["module"] = module + "({})".format(alias) if alias else module
```
vpodzime, line 513: Make sure the `if alias else module` actually applies to the part of the expression that you want it to apply.

cfbs/commands.py
```diff
@@ -460,5 +464,64 @@ def install_command(destination=None) -> int:
...
+            "version",
+            "status",
+            "by",
+            "tags",
+            "repo",
+            "commit",
+            "dependencies",
+            "added_by",
+            "description",
+        ]
```
vpodzime, line 512: I think this should be a tuple and it should be defined outside of this function and potentially reused in other places.

cfbs/commands.py
```diff
@@ -460,5 +464,64 @@ def install_command(destination=None) -> int:
...
+        data["module"] = module + "({})".format(alias) if alias else module
+        for key in ordered_keys:
+            if key in data:
+                if key in ["tags", "dependencies"]:
+                    value = " ".join(data[key])
```
vpodzime, line 517: Shouldn't this be `", ".join()`?

cfbs/main.py
```diff
@@ -85,6 +85,8 @@ def main() -> int:
         return commands.pretty_command(args.args, args.check)
     if args.command == "validate":
         return commands.validate_command(index=args.index)
+    if args.command in ["info", "show"]:
```
vpodzime, line 88: tuple here, please (nitpick)

vpodzime (commented):
Looks good to me otherwise. A nice addition!

cfbs/commands.py
craigcomstock: Oh. I did have it that way before I refactored to operate on dependencies as well. I liked it both ways but @nickanderson preferred commas.

cfbs/commands.py
```diff
@@ -459,3 +459,6 @@ def install_command(destination=None) -> int:
     rm(destination, missing_ok=True)
     cp("out/masterfiles", destination)
     return 0
+
+def help_command():
+    pass # no-op to allow help to be printed
```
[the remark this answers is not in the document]
craigcomstock: I think other parts of this mechanism could be improved but I really do like keeping it "simple" and leaving this function here. :) If you don't mind too much. I can add a better comment.

cfbs/commands.py
```diff
@@ -459,3 +462,65 @@ def install_command(destination=None) -> int:
...
+            found = any(m for m in build if m["name"] == module)
+            data["status"] = "Added" if found else "Not Added"
+        else:
+            if not module.startswith("./"):
+                module = "./" + module
+            data = next((m for m in build if m["name"] == module), None)
+            if data is None:
+                print("Path {} exists but is not yet added as a module.".format(module))
+                continue
+            data = data[0]  # take the first (and only) item
```
vpodzime, line 498: This is wrong now, AFAICT.

PR #29, opened 2021-10-14
Support adding modules from git repositories

vpodzime:
Aside from using the defaul/global index, it's handy to have an
ability to add modules from just a git repository. With a
'cfbs.json' file with the repository-specific index it can be
used as a source of modules.

Ticket: CFE-3771
Changelog: 'cfbs add' now supports adding modules from git repositories with index

TODO:
- [x] interactive *"Do you want to add all modules from the repo?"*
- [x] proper support for `ssh://git@example.com/some/git/repo/here`
- [x] support for `cfbs add https://some/repo@some_commit_sha`
- [x] add `index` info to `cfbs info`

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -568,6 +568,7 @@ def print_module_info(data):
         "by",
         "tags",
         "repo",
+        "index",
         "commit",
         "dependencies",
         "added_by",
@@ -591,17 +592,21 @@ def info_command(modules, index=None):
 
     for module in modules:
         print()  # whitespace for readability
-        if not index.exists(module):
+        in_build = any(m for m in build if m["name"] == module)
+        if not index.exists(module) and not in_build:
             print("Module '{}' does not exist".format(module))
             continue
-        if module in index:
+        if in_build:
+            # prefer information from the local source
+            data = next(m for m in build if m["name"] == module)
+            data["status"] = "Added"
+        elif module in index:
             data = index[module]
             if "alias" in data:
                 alias = module
                 module = data["alias"]
                 data = index[module]
-            found = any(m for m in build if m["name"] == module)
-            data["status"] = "Added" if found else "Not Added"
+            data["status"] = "Added" if in_build else "Not Added"
         else:
             if not module.startswith("./"):
                 module = "./" + module
--- a/test/test_showinfo.py
+++ b/test/test_showinfo.py
@@ -20,6 +20,7 @@ By: https:\/\/github.com\/cfengine
 Tags: wip, untested
 Repo: https:\/\/github.com\/cfengine\/modules
 Commit: [a-zA-Z0-9]+
+Added By: ./foo/main.cf
 Description: Enable autorun functionality
 """, out, re.M)
 


cfbs/commands.py
```diff
@@ -201,15 +202,96 @@ def local_module_copy(module, counter, max_length):
...
+    head_ref = head_ref_info[5:].strip()
+
+    with open(os.path.join(repo_path, ".git", head_ref)) as f:
+        return f.read().strip()
+
+def _clone_index_repo(repo_url):
+    assert repo_url.startswith(("https://", "ssh://", "git://"))
+
+    commit = None
+    if "@" in repo_url and (repo_url.rindex("@") > repo_url.rindex(".")):
```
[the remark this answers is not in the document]
vpodzime: Well, I think `@some-branch` would just work, right? But I don't quite want to support it because it might give people impression `cfbs build` would fetch/rebase from the branch. Better let them think they should use a SHA. :)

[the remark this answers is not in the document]
olehermanse: This is a valid point and I will create some tickets around it, starting with this one:

https://tracker.mender.io/browse/CFE-3797

(I'll add some more tickets to same epic).

No need to do anything more in this PR.

cfbs/commands.py
```diff
@@ -201,15 +202,96 @@ def local_module_copy(module, counter, max_length):
         f"{counter:03d} {pad_right(name, max_length)} @ local                                    (Copied)"
     )
 
+def _get_path_from_url(url):
+    if not url.startswith(("https://", "ssh://", "git://")):
+        if ("://" in url):
+            return user_error("Unsupported URL protocol in '%s'" % url)
+        else:
+            # It's a path already, just remove trailing slashes (if any).
+            return url.strip("/")
```
olehermanse, line 211: This would also strip a leading slash. Consider using rstrip, adding an assert / error message or at least updating the comment.

cfbs/commands.py
```diff
@@ -201,15 +202,96 @@ def local_module_copy(module, counter, max_length):
...
+
+    path = None
+    if url.startswith("ssh://"):
+        match = re.match(r"ssh://(\w+)@(.+)", url)
+        if match is not None:
+            path = match[2]
+    path = path or url[url.index("://") + 3:]
+    path = strip_right(path, ".git")
+    path = path.strip("/")
```
olehermanse, line 220: Again, `assert` somewhere and/or `rstrip`.

cfbs/commands.py
```diff
@@ -201,15 +202,96 @@ def local_module_copy(module, counter, max_length):
...
+            # Already cloned in the commit dir, just remove the 'master' clone
+            sh("rm -rf %s" % master_path)
+        else:
+            sh("mv %s %s" % (master_path, commit_path))
+
+    index_path = os.path.join(commit_path, "cfbs.json")
+    if os.path.exists(index_path):
+        return (index_path, commit)
+    else:
+        user_error("Repository '%s' doesn't contain a valid cfbs.json index file" % repo_url)
```
olehermanse, line 270: I see you are calling it an index file. That's not what I intended, up until now we've had 2 types of JSON files `cfbs.json` (config file for 1 project), `index.json` (index of all available modules).

BUT, what you did is smart and I think we should probably change this, standardize the format so there is only 1 format, `cfbs.json` and it can have an index inline or external path to index.

No need to do anything here, but let's discuss this some more :)

olehermanse (approved):
LGTM. I have some smaller nitpicks but feel free to merge it as is, and address things in follow-ups etc.

PR #34, opened 2021-10-26
CFE-3792: Make cfbs python 3.5 compatible

larsewi:
Ticket: CFE-3792
Changelog: None


olehermanse (approved):
Thanks!

PR #54, opened 2021-10-31
CFE-3767: Added command `cfbs remove`

larsewi:
Added command `cfbs remove` which can be used to remove added modules.


olehermanse (approved):
ACK; thanks.

PR #56, opened 2021-11-01
CFE-3821 Modified cfbs install behavior for easier use

craigcomstock:
If an argument is provided, install will use that path
If no argument and root is the user, default path of /var/cfengine/masterfiles is used
If no argument and NOT root user, $HOME/.cfagent/inputs is used

Ticket: CFE-3821


olehermanse (approved):
Makes sense, thanks!

PR #58, opened 2021-11-01
CFE-3829 Added error messages for missing required keys in module definition

craigcomstock:
Ticket: CFE-3829

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -589,7 +589,7 @@ def clean_command(non_interactive=False):
     modules = definition["build"]
 
     def _someone_needs_me(this) -> bool:
-        if "added_by" in this and this["added_by"] == "cfbs add":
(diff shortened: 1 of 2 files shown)


cfbs/commands.py
```diff
@@ -587,7 +589,7 @@ def clean_command(non_interactive=False):
     modules = definition["build"]
 
     def _someone_needs_me(this) -> bool:
-        if this["added_by"] == "cfbs add":
+        if "added_by" in this and this["added_by"] == "cfbs add":
```
olehermanse, line 592:
```suggestion
        if "added_by" not in this or this["added_by"] == "cfbs add":
```

olehermanse: I think the default here should be to say that "someone needs me" if added_by is missing.

PR #61, opened 2021-11-01
CFE-3830 Added error message in case of build property missing from cfbs.json

craigcomstock:
Refactored to catch more of build key missing.

Ticket: CFE-3830


olehermanse (approved):
Thanks.

PR #63, opened 2021-11-01
Added shell test for `cfbs clean`

larsewi:
Ticket: None
Changelog: None

### DIFF
--- a/test/shell/009_clean.sh
+++ b/test/shell/009_clean.sh
@@ -1,0 +1,15 @@
+set -e
+set -x
+cd test
+mkdir -p ./tmp/
+cd ./tmp/
+touch cfbs.json && rm cfbs.json
+
+cfbs init
+cfbs add promise-type-git
+grep '"name": "library-for-promise-types-in-python"' cfbs.json
+grep '"name": "promise-type-git"' cfbs.json
+cfbs remove promise-type-git --non-interactive
+cfbs clean
+! grep '"name": "library-for-promise-types-in-python"' cfbs.json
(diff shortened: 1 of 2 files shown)


test/shell/009_clean.sh
```diff
@@ -0,0 +1,15 @@
...
+mkdir -p ./tmp/
+cd ./tmp/
+touch cfbs.json && rm cfbs.json
+
+cfbs init
+cfbs add promise-type-git
+grep '"name": "library-for-promise-types-in-python"' cfbs.json
+grep '"name": "promise-type-git"' cfbs.json
+cfbs remove promise-type-git
+cfbs clean
```
olehermanse, line 13: I guess you should use `--non-interactive` here :)

test/shell/009_clean.sh
```diff
@@ -0,0 +1,15 @@
+set -e
```
olehermanse, line 1: Please add this to the `all.sh` script in the same folder.

olehermanse (changes requested):

craigcomstock (commented):
Agree with @olehermanse comments. Otherwise looks good.

olehermanse (changes requested):
Please rebase this now that the other one was merged. In hindsight it probably would have been better to do 2 commits in 1 PR in this case :)

olehermanse:
> Please rebase this now that the other one was merged. In hindsight it probably would have been better to do 2 commits in 1 PR in this case :)

Never mind, I fixed the conflict in GitHub UI :) Merging.

PR #64, opened 2021-11-02
CFE-3835 Added information about build step types to README

craigcomstock:
Ticket: CFE-3835


olehermanse (approved):
Great, thanks for working on this :)

PR #73, opened 2021-11-09
Added action to automatically format with Black

olehermanse:


olehermanse:
Merging to test GH Action.

PR #87, opened 2021-11-10
cfbs add will now print warnings when there are no autorun tags in policy files

olehermanse:
Like this:

```
$ cfbs add dir/one.cf
Added module: autorun (Dependency of ./dir/one.cf)
Added module: ./dir/one.cf
WARNING: No autorun tag found in policy file: './dir/one.cf'
WARNING: Tag the bundle(s) you want evaluated:
WARNING:   meta: "tags" slist => { "autorun" };
```


craigcomstock (approved):

cfbs/utils.py
```diff
@@ -237,3 +237,19 @@ def fetch_url(url, target, checksum=None):
...
+    assert os.path.isdir(name)
+    for root, subdirs, subfiles in os.walk(name):
+        if files:
+            for file in subfiles:
+                if not extension or (extension and file.endswith(extension)):
+                    yield os.path.join(root, file)
+        if directories:
+            for dir in subdirs:
+                if not extension or (extension and dir.endswith(extension)):
+                    yield os.path.join(root, dir) + "/"
```
vpodzime, line 253: I'd maybe yield directories first.

cfbs/index.py
```diff
@@ -235,6 +250,26 @@ def save(self, data=None):
...
+            if not _has_autorun_tag(name):
+                log.warning("No autorun tag found in policy file: '%s'" % name)
+                log.warning("Tag the bundle(s) you want evaluated:")
+                log.warning('  meta: "tags" slist => { "autorun" };')
+            return
+        if name.startswith("./") and name.endswith("/"):
+            assert os.path.isdir(name)
+            policy_files = list(find(name, extension=".cf"))
+            with_autorun = [x for x in policy_files if _has_autorun_tag(x)]
+            if len(policy_files) > 0 and len(with_autorun) == 0:
```
vpodzime, line 268: I'd do `if any(policy_files) and not any(with_autorun)` and used a generator instead of list for `with_autorun`, but YMMV.

cfbs/index.py
```diff
@@ -255,3 +290,4 @@ def add(self, module, remote_config=None, dependent=None):
             print("Added module: %s (Dependency of %s)" % (module["name"], dependent))
         else:
             print("Added module: %s" % module["name"])
+        self._validate_added_module(module)
```
vpodzime, line 293: Extra `_`, I think.

vpodzime (commented):
Looks good to me otherwise.

cfbs/index.py
```diff
@@ -235,6 +250,26 @@ def save(self, data=None):
...
         return module["name"] in (m["name"] for m in self["build"])
 
+    def validate_added_module(module):
+        """Try to help the user with warnings in appropriate cases"""
+
+        name = module["name"]
+        if name.startswith("./") and name.endswith(".cf"):
+            assert os.path.isfile(name)
+            if not _has_autorun_tag(name):
+                log.warning("No autorun tag found in policy file: '%s'" % name)
```
nickanderson, line 260: no current objection to these messages, I think it's clear. Possibly could be _more_ specific and say Tag the bundle(s) in FILENAME ...

nickanderson (approved):
LGTM, warnings are clear enough, my only potential suggestion makes the messages longer :)

cfbs/utils.py
olehermanse: Alright, no difference to me.

cfbs/index.py
olehermanse: Oki.

cfbs/index.py
olehermanse: The policy filename is there, in the first log message. I don't think I should repeat it.

vpodzime (approved):
ACK

PR #94, opened 2022-01-28
cfbs install to existing policy directory merges def.json instead of overwriting

craigcomstock:
also check if destination has a .git directory and bail

TODO: change this to a different command maybe? or an option `--merge-into` for install?

TODO: think through the use cases and possibilities and make this "just work" :)

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -628,7 +628,7 @@ def build_command() -> int:
     build_steps(config)
 
 
-def install_command(args) -> int:
+def install_command(args, merge=False) -> int:
     if len(args) > 1:
         user_error(
             "Only one destination is allowed for command: cfbs install [destination]"
@@ -642,14 +642,27 @@ def install_command(args) -> int:
         destination = "/var/cfengine/masterfiles"
     if len(args) > 0:
         destination = args[0]
+        merge = True
     elif os.getuid() == 0:
         destination = "/var/cfengine/masterfiles"
     else:
         destination = os.path.join(os.environ["HOME"], ".cfagent/inputs")
     if not destination.startswith("/") and not destination.startswith("./"):
         destination = "./" + destination
-    rm(destination, missing_ok=True)
-    cp("out/masterfiles", destination)
+    if not merge and os.path.exists(os.path.join(destination, ".git")):
+        user_error( "destination, {}, has a .git directory, not installing. Destination should be a previously installed directory or non-existent." )
+
+    print( "destination is {}, merge is {}".format(destination, merge))
+#    return 0
+    if merge:
+      built, original = read_json("out/masterfiles/def.json"), read_json(os.path.join(destination, "def.json"))
+      merged = merge_json(original, built)
+      write_json(os.path.join(destination, "def.json.new"), merged)
+      cp("out/masterfiles", destination)
+      cp(os.path.join(destination, "def.json.new"), os.path.join(destination, "def.json"))
+    else:
+      rm(destination, missing_ok=True)
+      cp("out/masterfiles", destination)
     print("Installed to %s" % destination)
     return 0
 
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -628,7 +628,7 @@ def build_command() -> int:
     build_steps(config)
 
 
-def install_command(args, merge=False) -> int:
+def install_command(args, force=False) -> int:
     if len(args) > 1:
         user_error(
             "Only one destination is allowed for command: cfbs install [destination]"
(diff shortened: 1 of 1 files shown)


nickanderson:
So, the current usage here is:

`cfbs install PATH`?

If PATH to install to is provided then merge def.json instead of overwriting it.

Also, in current implementation, this does not work with respect to removing something.

E.g. i `cfbs add tmp-file-age inventory-etc-login-defs` in a new cfbs project. I `cfbs build` and `cfsb install /my/masterfiles`. Up to this point things work as expected, assets from both modules are added to /my/masterfiels and the def.json from the build is merged into whatever I already have. Now I `cfbs remove tmp-file-age; cfbs build; cfbs install /my/masterfiles`. The result is an updated def.json that does not include the bits related to the `tmp-file-age` module however the artifacts from the tmp-file-age module that were placed previously are left undisturbed.

I am not sure if it's worth having this ability to merge json on top during install without some support for cleanup. Thoughts @olehermanse , @vpodzime, @craigcomstock ?

cfbs/commands.py
```diff
@@ -642,14 +642,27 @@ def install_command(args) -> int:
         destination = "/var/cfengine/masterfiles"
     if len(args) > 0:
         destination = args[0]
+        merge = True
```
vpodzime, line 645: I think this should be based on an option plus an interactive prompt.

olehermanse: Yea, this needs to be an option.

olehermanse (changes requested):
This should not happen by default, and I'm not sure why it's being added at all? It goes against the expectation of deterministic results, `cfbs build && cfbs install` on 2 different hubs should result in the same policy set. Please expand the commit message with some information, and add it as an option that has to be explicitly used.

craigcomstock:
> This should not happen by default, and I'm not sure why it's being added at all? It goes against the expectation of deterministic results, `cfbs build && cfbs install` on 2 different hubs should result in the same policy set. Please expand the commit message with some information, and add it as an option that has to be explicitly used.

It was added as a WIP to show how easy the functionality was available. I expected some discussion about exactly WHEN it should occur, which has happened above.

I do think in some circumstances it should "just work" and indeed merge. But as Nick pointed out, this doesn't work when modules are removed so some other mechanism might be needed in that case, such as saving off a copy of the target masterfiles def.json to be used instead of the built version. Then again, this is getting into "not easy" due to trying to support another use case: installing into an existing policy set. Maybe we should just stick to improving one use case: import your policy set and manage with cfbs.

vpodzime:
I feel like a better strategy would be to use the `augments` key and implementing merging of augments data when it's being loaded.

craigcomstock:
> I feel like a better strategy would be to use the `augments` key and implementing merging of augments data when it's being loaded.

You like there being two ways to install then? 1) install to fresh dir aka wipe target out and re-install and 2) install into existing policy set?

And then in the second case sending over the cfbs augments as a separate json file and only adding the augments key in the target def.json file as a way of integrating cfbs augments data?
This would handle @nickanderson 's case of removing a module in cfbs. This makes sense to me.

Does the augments key loading a separate json file work? As in the vars/classes/etc are scoped the same way regardless of if they are in def.json or an augments key loaded json file?

Reading here, https://docs.cfengine.com/docs/3.18/reference-language-concepts-augments.html#augments, it would seem to do the right thing.

So I would agree that if we want to support installing into an existing policy dir we should provide this as an option. @vpodzime  how would you like this to look? Below are some thoughts on how I would like it to work:

```
cfbs install my/mpf # if def.json present then send cfbs augments as def-cfbs.json and add augments key to existing def.json
cfbs install my/mpf # if no def.json then just write cfbs augments as def.json
```

Another case I was hoping to address with this was if the target had a .git dir, interactively ask if overwrite is ok or provide an option to `--force`?

olehermanse:
@craigcomstock The default for cfbs is to build and install a complete, deterministic policy set, not affected by what happens to be on the host where it's installed. The policy set is everything in `/var/cfengine/masterfiles`. Data is separate, and can be handled in many other ways, including CMDB, custom modules, or multiple augments files.

There should be no difference in behavior between specifying install dir and not:

```
cfbs install
cfbs install /var/cfengine/masterfiles
```

If you think the merge functionality is useful, please make it an option:

```
cfbs install --merge
cfbs install --merge /var/cfengine/masterfiles
```

So it's not an unexpected (unpleasant) surprise.

craigcomstock:
On further reflection I think we should only focus on and support installing to something like /var/cfengine/masterfiles which is not under git control. So add features to cfbs to enable integrating existing policy sets into the cfbs workflow instead of trying to add them on to the side. I'll rework this PR to add a `--force` which will allow installing to a directory which has an existing `.git` dir. This is a safety issue for folks who have /var/cfengine/masterfiles as their git policy clone directory. They might lose work if we don't check first.

vpodzime:
> @vpodzime how would you like this to look?

IMHO, all modules should ship their *def.json* which should be added to the `augments` key of the top-level *def.json*. But we need the merging functionality first.

PR #95, opened 2022-02-02
Fixed bugs with adding URL at specific commit

olehermanse:
- Fixed bug with adding URL at specific commit
- Added test for adding a URL at specific commit
- Fixed problems with adding/cloning the same URL and commit twice

### DIFF
--- a/cfbs/core.py
+++ b/cfbs/core.py
@@ -157,7 +157,8 @@ class Index:
         return module
 
 
-def _construct_provided_module(name, data, url, url_commit):
+def _construct_provided_module(name, data, url, commit):
+    assert "@" not in url  # Should already be split into url and commit
     module = OrderedDict()
     module["name"] = name
     if "description" not in data:
@@ -166,7 +167,7 @@ def _construct_provided_module(name, data, url, url_commit):
         )
     module["description"] = data["description"]
     module["url"] = url
-    module["commit"] = url_commit
+    module["commit"] = commit
     subdirectory = data.get("subdirectory")
     if subdirectory:
         module["subdirectory"] = subdirectory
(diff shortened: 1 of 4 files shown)


craigcomstock (approved):

cfbs/core.py
```diff
@@ -329,6 +330,10 @@ def _add_using_url(
             assert url.startswith(("https://", "git://", "ssh://"))
             config_path, url_commit = clone_url_repo(url)
 
+        if "@" in url:
+            assert url.split("@")[1] == url_commit
+            url = url.split("@")[0]
```
vpodzime, line 335: This needs to be a bit more careful. This is the code I wrote when I was originally adding support for modules from git repos:
```
    if "@" in repo_url and (repo_url.rindex("@") > repo_url.rindex(".")):
        # commit specified in the url
        repo_url, commit = repo_url.rsplit("@", 1)
```
The reason for the complexity is that `user@some-URL` can appear in the URLs, so splitting on `@` must be done more carefully.

cfbs/internal_file_management.py
```diff
@@ -139,6 +139,13 @@ def _get_git_repo_commit_sha(repo_path):
         return f.read().strip()
 
 
+def _clone_and_checkout(url, path, commit):
+    # NOTE: If any of these shell (git) commands fail, we will exit
+    if not os.path.exists(path + "/.git"):
```
vpodzime, line 144: Please use `os.path.join()`.

olehermanse: Fixed, thanks!

cfbs/core.py
olehermanse: Thank you, fixed. See last 2 commits.

PR #96, opened 2022-02-04
CFE-3904: The validate command no longer does head requests

larsewi:
The validate command is used in a GHA and it is making too many head
requests. Thus we'll remote them from the validate command, followed by
adding them in another GHA that only does head requests if the URL of a
remote resource is changed.

The GHA that will validate remote resources: https://github.com/cfengine/build-index/pull/225

Ticket: CFE-3904
Changelog: None
Signed-off-by: larsewi <lars.erik.wik@northern.tech>

### DIFF
--- a/cfbs/validate.py
+++ b/cfbs/validate.py
@@ -1,7 +1,6 @@
 import argparse
 import json
 import sys
-import requests
 import re
 
 from cfbs.utils import is_a_commit_hash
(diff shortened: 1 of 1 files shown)


olehermanse (approved):

vpodzime (approved):
ACK

PR #101, opened 2022-02-23
CFE-3846: Now you can specify version when adding modules

larsewi:


cfbs/core.py
```diff
@@ -123,36 +127,51 @@ def data(self) -> dict:
...
+    def get_module_object(self, name, version="latest", added_by=None):
         module = OrderedDict({"name": name})
         if name.startswith("./"):
             object = _generate_local_module_object(name)
         else:
             object = self[name]
+            if version != "latest":
+                # Change fields according to versions.json
+                versions = get_json(_VERSION_INDEX)
+                object.update({"version": version})
```
olehermanse, line 171:
```suggestion
                object["version"] = version
```

cfbs/core.py
```diff
@@ -123,36 +127,51 @@ def data(self) -> dict:
...
+                for key, val in versions[name][version].items():
+                    if key in object:
+                        object.update({key: val})
```
olehermanse, line 174:
```suggestion
                new_version = versions[name][version]
                object.update({"version": version})
                desired = ["commit", "version", "steps", "subdirectory", "by", "repo", "dependencies"]
                new_values = {k,v in new_version.items() if k in desired}
                object.update(new_values)
```

vpodzime (commented):
Looks good to me otherwise.

cfbs/cfbs_config.py
```diff
@@ -145,27 +147,26 @@ def _convert_added_by(added_by, to_add):
...
 
         return added_by
 
     def _update_added_by(self, requested, added_by):
-        for module in self["build"]:
-            if module["name"] in requested:
-                new_added_by = added_by[module["name"]]
-                if new_added_by == "cfbs add":
-                    module["added_by"] = "cfbs add"
+        for req, mod in itertools.product(requested, self["build"]):
```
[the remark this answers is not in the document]
larsewi: You're right, I'll fix!

cfbs/cfbs_config.py
```diff
@@ -208,30 +208,30 @@ def _add_without_dependencies(self, modules):
 
     def _add_modules(
         self,
-        names: list,
+        to_add: list,
```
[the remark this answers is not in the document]
larsewi: Good catch 👍

cfbs/module.py
```diff
@@ -0,0 +1,53 @@
...
+        assert key in Module.allowed_keys()
+        return super().get(key, default)
+
+    def update(self, m):
+        for key in m.keys():
+            assert key in Module.allowed_keys()
+        super().update(m)
+
+    @staticmethod
+    def allowed_keys():
```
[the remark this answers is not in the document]
larsewi: > Can we use a traditional class with attributes?

Traditional class is a bit wonky in that some attributes are optional while others are mutually exclusive.

> Can we solve this with properties?

Yes, but it requires a lot of code for each attribute. Therefore I used the `__getattr__` and `__setattr__` functions instead to implement functionality for all attributes at once.

cfbs/cfbs_config.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import os
 import logging as log
+import itertools
```
olehermanse, line 4: Unused import

test/shell/014_add_nonexistent.sh
```diff
@@ -7,4 +7,4 @@ touch cfbs.json && rm cfbs.json
 
 cfbs init
 ! ( cfbs add bollocks > output.log 2>&1 )
-grep -F "Error: Module 'bollocks' does not exist" output.log
+grep -F "Error: Module 'bollocks' version 'latest' does not exist" output.log
```
olehermanse, line 10: IMO, this error message was nicer before and we could make it only print version in case it's not latest.

cfbs/index.py
```diff
@@ -0,0 +1,161 @@
...
+            name = module
+            version = None
+        if os.path.exists(name):
+            return True
+        else:
+            if not version:
+                return name in self
+            else:
+                versions = get_json(_VERSION_INDEX)
+                return name in versions and version in versions[name]
```
olehermanse, line 114: Early return makes these else's unnecessary:

```suggestion
        if os.path.exists(name):
            return True
        if not version:
            return name in self
        versions = get_json(_VERSION_INDEX)
        return name in versions and version in versions[name]
```

https://refactoring.guru/replace-nested-conditional-with-guard-clauses

olehermanse (approved):
Looks good, thanks for the improvements and adding tests :)

I left some comments but feel free to address them in a follow-up.

cfbs/cfbs_config.py
```diff
@@ -145,27 +147,27 @@ def _convert_added_by(added_by, to_add):
...
         return added_by
 
     def _update_added_by(self, requested, added_by):
-        for module in self["build"]:
-            if module["name"] in requested:
-                new_added_by = added_by[module["name"]]
-                if new_added_by == "cfbs add":
-                    module["added_by"] = "cfbs add"
+        for req in requested:
+            for mod in self["build"]:
```
vpodzime, line 159: Would it be possible to do `if req["name"] in self["build"]:`?

cfbs/module.py
```diff
@@ -0,0 +1,74 @@
+from collections import OrderedDict
+
+from cfbs.pretty import pretty
+
+
+class Module(OrderedDict):
```
vpodzime, line 6: Please double-check the methods in the picture are OK for `Module` objects:
image

vpodzime (commented):
Looks good to me otherwise.

cfbs/module.py
vpodzime: Also, what about `del module["name"]`?

cfbs/cfbs_config.py
olehermanse: `self["build"]` is a list (JSON array) so this doesn't work, but we could add a helper, `self.already_added(name)`.

cfbs/module.py
```diff
@@ -0,0 +1,68 @@
...
+                return None
+            raise e
+
+    def __str__(self) -> str:
+        return pretty(self.to_dict())
+
+    def to_dict(self):
+        keys = Module.attributes()
+        vals = [getattr(self, attr) for attr in keys]
+        return OrderedDict({key: val for (key, val) in zip(keys, vals) if val})
```
vpodzime, line 50: You are creating an OrderedDict from a normal dict here. That doesn't make much sense. I think you need to use the tuples directly.

cfbs/index.py
```diff
@@ -97,35 +98,65 @@ def data(self) -> dict:
...
-        module = OrderedDict({"name": name})
         if name.startswith("./"):
             object = _generate_local_module_object(name)
         else:
             object = self[name]
+            if version:
+                versions = get_json(_VERSION_INDEX)
+                new_values = versions[name][version]
+                specifics = {
+                    k: v for (k, v) in new_values.items() if k in Module.keys()
```
vpodzime, line 156: This should be `Module.attributes()`, right?

vpodzime (commented):
Looks good to me. Although it's a bit of a mystery to me how the tests can pass with the `Module.keys()` bug.

vpodzime:
> Although it's a bit of a mystery to me how the tests can pass with the Module.keys() bug.

Oh, mystery solved -- the don't pass. Only on the commit adding the functionality because the test is in the commit following it.

cfbs/index.py
larsewi: Good catch 🙌

cfbs/cfbs_config.py
larsewi: Fixed

test/shell/014_add_nonexistent.sh
larsewi: Agreed 👍

cfbs/module.py
larsewi: N/A

vpodzime (approved):
Let's not waste any extra time on this.

PR #103, opened 2022-03-21
CFE-3813: Track changes done by cfbs commands in Git

vpodzime:


olehermanse (approved):
Looks very good in general, thanks for working on this! 🚀 I am sure I will find some things we might want to change with the usability / prompts, but we can take that as we start using it :)

PR #104, opened 2022-04-01
Fixed local install version numbers

olehermanse:
pip no longer accepted the version numbers straight from git,
so I had to convert it.


larsewi (approved):
ACK

PR #107, opened 2022-04-19
Handle EOF in prompts

vpodzime:
Same as if user just hit ENTER -- use the default.

Ticket: ENT-8488
Changelog: None

### DIFF
--- a/cfbs/prompts.py
+++ b/cfbs/prompts.py
@@ -28,7 +28,11 @@ def prompt_user(prompt, choices=None, default=None):
 
     answer = None
     while answer is None:
-        answer = input(prompt)
+        try:
+            answer = input(prompt)
+        except EOFError:
+            answer = ""
+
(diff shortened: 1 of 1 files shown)


cfbs/prompts.py
```diff
@@ -28,7 +28,11 @@ def prompt_user(prompt, choices=None, default=None):
...
     answer = None
     while answer is None:
-        answer = input(prompt)
+        try:
+            answer = input(prompt)
+        except EOFError:
+            answer = ""
+
         if answer == "":
             if default is not None:
```
larsewi, line 37: Isn't
```python
if default is not None:
    answer = default
else:
    answer = None
```
equivalent to 
```python
answer = default
```
?

larsewi (approved):
ACK

olehermanse (approved):

PR #108, opened 2022-04-21
Remove unnecessary if else condition in promts.py

larsewi:
- Add *.swp files to .gitignore
- Removed unnecessary if else condition in promts.py


vpodzime (approved):
ACK. Thanks! :sweat_smile:

olehermanse (approved):

PR #109, opened 2022-04-25
Remove the dependency on the 'requests' Python package

vpodzime:
The 'requests' package is not part of the standard Python library
and build set. We don't use the package extensively so to avoid
requiring dependencies, we can just use 'urllib' which is part of
the standard library and build set.

Ticket: ENT-8439
Changelog: Removed dependency on the 'requests' package


olehermanse (approved):
LGTM, thanks :)

PR #110, opened 2022-04-27
CFE-3812: cfbs init now prompts user whether to use default masterfiles

larsewi:
The cfbs init command now prompts users whether or not they want to use
the default masterfiles policy framework. If the user says no, they're
prompted with which masterfiles they want to add, allowing them to skip
if empty.

Ticket: CFE-3812
Changelog: Title

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -210,7 +210,7 @@ def init_command(index=None, non_interactive=False) -> int:
         to_add = "masterfiles"
     else:
         to_add = prompt_user(
-            "Specify policy set to use instead (or skip if empty)", default=""
+            "Specify policy set to use instead (empty to skip)", default=""
         )
 
     if to_add:
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -127,7 +127,7 @@ def init_command(index=None, non_interactive=False) -> int:
         "name": name,
         "type": "policy-set",  # TODO: Prompt whether user wants to make a module
         "description": description,
-        "build": [],  # TODO: Prompt what masterfile user wants to add
+        "build": [],
     }
     if index:
         config["index"] = index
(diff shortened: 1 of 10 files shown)


vpodzime (commented):
So `cfbs --non-interactive init` now adds MPF? Is that what we want?

cfbs/commands.py
```diff
@@ -193,7 +193,26 @@ def init_command(index=None, non_interactive=False) -> int:
...
+    CFBSConfig.instance = None
+
+    if prompt_user(
+        "Do you wish to add default masterfiles?",
+        choices=YES_NO_CHOICES,
+        default="yes",
+    ) in ("yes", "y"):
+        to_add = "masterfiles"
+    else:
+        to_add = prompt_user("Specify masterfiles module or skip if empty", default="")
```
larsewi, line 212:
```suggestion
        to_add = prompt_user("Specify policy set to use instead (or skip if empty)", default="")
```

larsewi:
> So `cfbs --non-interactive init` now adds MPF? Is that what we want?

The option `--non-interactive` should probably perform the default behaviour?  In this case the question would be; should `cfbs init` add MPF by default? 🤔  What do you think @vpodzime @olehermanse?

olehermanse:
> So `cfbs --non-interactive init` now adds MPF? Is that what we want?

Yes

olehermanse:
> > So `cfbs --non-interactive init` now adds MPF? Is that what we want?
> 
> 
> 
> The option `--non-interactive` should probably perform the default behaviour?  In this case the question would be; should `cfbs init` add MPF by default? 🤔

Yes and yes

olehermanse (changes requested):

olehermanse (approved):

PR #111, opened 2022-04-29
ENT-8573 & ENT-8579: cfbs update now supports specifying modules

larsewi:
cfbs update now supports specifying module name (or name and version) by
using the following format `cfbs update [NAME[@VERSION]]...`.

Ticket: ENT-8573 and ENT-8579
Changelog: Body

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -44,6 +44,7 @@ from cfbs.git import (
 )
 from cfbs.git_magic import commit_after_command, git_commit_maybe_prompt
 from cfbs.prompts import YES_NO_CHOICES, prompt_user
+from cfbs.module import Module
 
 
 _MODULES_URL = "https://archive.build.cfengine.com/modules"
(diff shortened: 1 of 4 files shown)


cfbs/commands.py
```diff
@@ -401,19 +402,47 @@ def _someone_needs_me(this) -> bool:
     return 0
 
 
-@commit_after_command("Updated all modules")
-def update_command():
-    new_deps = []
-    new_deps_added_by = dict()
+def update_command(to_update):
```
vpodzime, line 405: I'd prefer keeping the decorator in place. It guarantees that the code in it will run when the function returns. Having the code at the end of the function means every early `return` will skip it. We may need to extend the decorator to support getting the arguments for the commit message from the return tuple of the decorated function. So the function would do `return (0, list(updated_modules))` and the decorator would use that for the commit message.

cfbs/commands.py
```diff
@@ -401,19 +402,47 @@ def _someone_needs_me(this) -> bool:
...
     config = CFBSConfig.get_instance()
     index = config.index
+
+    if to_update:
+        to_update = [Module(m) for m in to_update]
+        index.translate_aliases(to_update)
+        skip = (
+            m for m in to_update if all(n["name"] != m.name for n in config["build"])
```
vpodzime, line 413: I'd use `not any(...==..)` because it reads better to me than `all(...!=...)`, but YMMV.

cfbs/commands.py
```diff
@@ -401,19 +402,47 @@ def _someone_needs_me(this) -> bool:
...
+    new_deps = []
+    new_deps_added_by = dict()
     changes_made = False
+    updated = []
+
     for module in config["build"]:
+        update = None
+        for m in to_update:
+            if m.name == module["name"]:
+                update = m
```
vpodzime, line 431: This logic where there's a list of things to update, but then the iteration goes over all modules instead of the things to update is a bit twisted, IMHO. I'd expect something like this:
```
for update_req in to_update:
    name = get_name_from_update_req(update_req)
    module = config["build"].get(name)
    ...
```

cfbs/commands.py
```diff
@@ -444,15 +473,19 @@ def update_command():
...
+                log.warning(
+                    "The requested version of module '%s' is older than current version. (%s < %s)",
+                    module["name"],
+                    index_info["version"],
+                    module["version"],
+                )
+                if prompt_user(
+                    "Do you wish to downgrade?", choices=YES_NO_CHOICES, default="yes"
+                ) in ("no", "n"):
+                    continue
```
vpodzime, line 488: Please don't allow downgrades.

cfbs/commands.py
```diff
@@ -495,12 +528,25 @@ def update_command():
...
+        if not update.version:
+            update.version = index_info["version"]
+        updated.append(update)
+
     if new_deps:
         objects = [index.get_module_object(d, new_deps_added_by[d]) for d in new_deps]
         config.add_with_dependencies(objects)
     config.save()
 
-    return 0 if changes_made else 2
```
vpodzime, line 503: Ole introduced a special exception for no changes.

vpodzime (commented):
Looks good to me otherwise.

cfbs/commands.py
olehermanse: Not really a strong preference, but I think I prefer `all`

cfbs/commands.py
vpodzime: And make it be a `namedtuple`, maybe? I don't quite care, TBH.

olehermanse (approved):
Merging as the main issue has been resolved.

vpodzime:
> Merging as the main issue has been resolved.

Fair enough, but my comments are all still applicable.

PR #112, opened 2022-05-03
CFE-3905: Adjust version parsing to include dashes as well as periods

craigcomstock:
Ticket: CFE-3905
Changelog: title


cf-bottom:
Thank you for submitting a PR! Maybe @vpodzime can review this?

cfbs/validate.py
```diff
@@ -84,7 +84,7 @@ def validate_version(name, modules):
             raise CFBSIndexException(name, "Missing required attribute 'version'")
         if type(modules[name]["version"]) != str:
             raise CFBSIndexException(name, "'version' must be of type string")
-        regex = r"(0|[1-9][0-9]*).(0|[1-9][0-9]*).(0|[1-9][0-9]*)"
+        regex = r"(0|[1-9][0-9]*).(0|[1-9][0-9]*).(0|[1-9][0-9]*)(-([0-9]*)?)?"
```
vpodzime, line 87: The `*` should be `+` and the first `?` should be dropped, AFAICT. We don't want versions like `1.0.0-`, there must be at least one digit.

vpodzime (commented):
The second commit doesn't have a ticket reference. Looks good to me other than that and the minor issue in the regex.

cfbs/validate.py
craigcomstock: sure, makes sense.

craigcomstock: I also converted the `.` to `\.` so that you something like `3.1-20` doesn't validate. :)

vpodzime (approved):
ACK

PR #115, opened 2022-05-24
Fixed inconsistency when handling local folders as modules

olehermanse:
- Fixed inconsistency when handling local folders as modules
- README.md: Added some more explanation about build steps
- Ensured cfbs init respects --non-interactive


vpodzime (approved):
ACK

PR #117, opened 2022-06-03
Follow-ups on ENT-8573 & ENT-8579

larsewi:
Follow-ups from https://github.com/cfengine/cfbs/pull/111

With the new Result namedtuple we can override the commit message during command execution.

Output with one update:
```
$ cfbs update
The default commit message is 'Update module 'promise-type-groups' to version 0.1.3' - edit it? [yes/y/NO/n] 
Committing using git:

[master 8a71ec6] Update module 'promise-type-groups' to version 0.1.3
 1 file changed, 2 insertions(+), 2 deletions(-)

```

Output with more than one update:
```
$ cfbs update
The default commit message is:

	Updated 2 modules

	 - Updated module 'promise-type-groups' to version 0.1.3
	 - Updated module 'promise-type-git' to version 0.1.1

Edit it? [yes/y/NO/n] 
Committing using git:

[master b2d3ea8] Updated 2 modules
 1 file changed, 4 insertions(+), 4 deletions(-)

```

### DIFF
--- a/cfbs/git_magic.py
+++ b/cfbs/git_magic.py
@@ -4,10 +4,14 @@ Here it's okay to depend on other parts of the CFBS codebase,
 do prompts, etc.
 """
 
+from collections import namedtuple
 from cfbs.prompts import YES_NO_CHOICES, prompt_user
 from cfbs.cfbs_config import CFBSConfig, CFBSReturnWithoutCommit
 from cfbs.git import git_commit, git_discard_changes_in_file, CFBSGitError
 from functools import partial
+
+
+Result = namedtuple("Result", ["rc", "commit", "msg"])
 
 
 def git_commit_maybe_prompt(commit_msg, non_interactive, scope="all"):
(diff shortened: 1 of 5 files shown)


cfbs/git_magic.py
```diff
@@ -4,12 +4,16 @@
...
 """
 
+from collections import namedtuple
 from cfbs.prompts import YES_NO_CHOICES, prompt_user
 from cfbs.cfbs_config import CFBSConfig, CFBSReturnWithoutCommit
 from cfbs.git import git_commit, git_discard_changes_in_file, CFBSGitError
 from functools import partial
 
 
+Result = namedtuple("Result", ["rc", "commit", "msg_body"])
```
olehermanse, line 14: I don't like that you can only set the body, but not the title (first line) of the commit message in the result. I'd want to be able to create different commit messages (including title) depending on what happened in the function.

olehermanse (changes requested):
Looks good in general, but I think we should enable setting the whole commit message in the result.

cfbs/git_magic.py
larsewi: Now you override the entire commit message if 
```Python
msg is not None
```

olehermanse (approved):
LGTM

vpodzime (commented):
Very nice! :clap:

PR #119, opened 2022-06-08
ENT-8585: Added command line options for setting / overriding the git details used for commits

larsewi:
Option to override git:
 - [x] attribute in cfbs.json
 - [x] user name
 - [x] user email
 - [x] commit message

### DIFF
--- a/cfbs/args.py
+++ b/cfbs/args.py
@@ -1,0 +1,65 @@
+import argparse
+
+from cfbs import commands
+from cfbs.utils import cache
+
+
+def get_args():
+    parser = _get_arg_parser()
+    args = parser.parse_args()
+    return args
+
+
+def print_help():
+    parser = _get_arg_parser()
+    parser.print_help()
+
+
+@cache
+def _get_arg_parser():
+    command_list = [
+        cmd.split("_")[0] for cmd in dir(commands) if cmd.endswith("_command")
+    ]
+    parser = argparse.ArgumentParser(description="CFEngine Build System.")
+    parser.add_argument(
+        "command",
+        metavar="cmd",
+        type=str,
+        nargs="?",
+        help="The command to perform ({})".format(", ".join(command_list)),
+    )
+    parser.add_argument("args", nargs="*", help="Command arguments")
+    parser.add_argument(
+        "--loglevel",
+        "-l",
+        help="Set log level for more/less detailed output",
+        type=str,
+        default="warning",
+    )
+    parser.add_argument(
+        "--version", "-V", help="Print version number", action="store_true"
+    )
+    parser.add_argument(
+        "--force", help="Force rebuild / redownload", action="store_true"
+    )
+    parser.add_argument(
+        "--non-interactive",
+        help="Don't prompt, use defaults (only for testing)",
+        action="store_true",
+    )
+    parser.add_argument("--index", help="Specify alternate index", type=str)
+    parser.add_argument(
+        "--check", help="Check if file(s) would be reformatted", action="store_true"
+    )
+    parser.add_argument(
+        "--checksum",
+        type=str,
+        default=None,
+        help="Expected checksum of the downloaded file",
+    )
+    parser.add_argument(
+        "--keep-order",
+        help="Keep order of items in the JSON in 'cfbs pretty'",
+        action="store_true",
+    )
+    return parser
--- a/cfbs/args.py
+++ b/cfbs/args.py
@@ -67,4 +67,12 @@ def _get_arg_parser():
         choices=("yes", "no"),
         help="Override git option in cfbs.json",
     )
+    parser.add_argument(
+        "--git-user-name",
+        help="Specify git user name",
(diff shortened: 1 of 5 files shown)


cfbs/main.py
```diff
@@ -127,7 +127,7 @@ def main() -> int:
         _get_arg_parser().print_help()
         return 0
 
-    config = CFBSConfig.get_instance(args.index, args.non_interactive)
+    _ = CFBSConfig.get_instance(args.index, args.non_interactive)
```
vpodzime, line 74: We don't even need to assign the instance to a variable, right?

cfbs/args.py
```diff
@@ -62,4 +62,9 @@ def _get_arg_parser():
         help="Keep order of items in the JSON in 'cfbs pretty'",
         action="store_true",
     )
+    parser.add_argument(
+        "--git",
+        choices=("yes", "no"),
+        help="Override git option in cfbs.json",
```
vpodzime, line 68: Isn't there something like `type=bool` that understands the usual "boolean" values for options and sets the value to `None, True, False` based on it?

cfbs/git.py
```diff
@@ -85,14 +85,18 @@ def git_init(user_name=None, user_email=None, description=None):
...
+    commit_msg, edit_commit_msg=False, user_name=None, user_email=None, scope="all"
+):
     """Create a commit in the CWD Git repository
 
     :param commit_msg: commit message to use for the commit
     :param scope: files to include in the commit or `"all"` (`git commit -a`)
     :type scope: str or an iterable of str
     :param edit_commit_message=False: whether the user should be prompted to edit and
                             save the commit message or not
+    :param user_name=None: override git config user name
```
vpodzime, line 98: I don't think the `=None` part is necessary here (and maybe even not supported by Sphinx?).

cfbs/git_magic.py
```diff
@@ -4,20 +4,35 @@
 do prompts, etc.
 """
 
+import logging as log
```
vpodzime, line 7: This adds a duplicate `import` already done below.

cfbs/git_magic.py
```diff
@@ -4,20 +4,35 @@
...
+commit_counter = counter(start=1)
+
 
 def git_commit_maybe_prompt(commit_msg, non_interactive, scope="all"):
     edit_commit_msg = False
+    args = get_args()
+
+    # Override message if --git-commit-message option is used
+    if args.git_commit_message:
+        if commit_counter.__next__() > 1:
```
vpodzime, line 29: This could be replaced with a simple boolean value `first_commit = True`.

vpodzime (commented):
Looks good to me otherwise.

cfbs/main.py
larsewi: It is needed for tests to pass, but I haven't investigated why

cfbs/args.py
larsewi: There is a `BooleanOptionalAction` that accepts `--git` or `--no-git`, but it was not introduced before python 3.9. I could of course subclass `Action` and create my own if that would be desirable.

cfbs/main.py
vpodzime: That's weird, it shouldn't be needed.

cfbs/args.py
vpodzime: That's probably not necessary.

cfbs/main.py
larsewi: Here we create the instance with arguments
```python3
_ = CFBSConfig.get_instance(args.index, args.non_interactive)
```

All other places we just get the already created instance *_(as it is a singleton)_* and don't pass any arguments
```python3
config = CFBSConfig.get_instance()
```
Thus we need it 😉

vpodzime: So what exactly fails and how? I don't think you are required to assign the returned value to a variable in Python if you don't care about it.

larsewi: I.e. the tests halt because the "--non-interactive" option is not passed to the config.

vpodzime (approved):
Looks good to me now.

cfbs/commands.py
```diff
@@ -133,22 +134,27 @@ def init_command(index=None, non_interactive=False) -> int:
     if index:
         config["index"] = index
 
+    do_git = get_args().git
```
olehermanse, line 138: Here and below you are effectively introducing another global, another "hidden dependency" for this function. I don't like this - IMO these things should be a part of the config, `CFBSConfig` is the current config of the appliation, meaning JSON + command line options. I think it would be useful if we follow some patterns here, for example that the calling a `_command` functions with the right arguments is equivalent to running the cfbs command.

I can do some refactoring and add some comments to make this more clear.

olehermanse (approved):
There are some things I'd like to change structuring this, but they are bigger clarifications / refactorings, so I'll just submit a follow-up.

PR #122, opened 2022-07-12
CFE-3953: added user input support for cfbs

larsewi:
Here are some example modules to test with [cfbs-test-modules](https://github.com/larsewi/cfbs-test-modules)
 - [x] Write documentation
 - [x] Implement `cfbs input` command
 - [x] Remove input data on `cfbs remove`
 - [x] Prompt for input on `cfbs add`
 - [x] Update input on `cfbs update`
 - [x] Create augments on `cfbs build`


cf-bottom:
Thanks for submitting a pull request! Maybe @olehermanse can review this?

cfbs/cfbs_config.py
```diff
@@ -298,9 +299,15 @@ def add_command(
...
-            # Not an error, we just want to exit successfully without
-            # making a git commit
-            raise CFBSReturnWithoutCommit(0)
-
-        return 0
+        msg = ""
+        count = 0
+        for module in self["build"]:
+            if not any(module["name"] == m["name"] for m in before):
+                msg += "\n - Added module '%s'" % module["name"]
```
vpodzime, line 344: I'd prefer a set difference to be used here, but YMMV.

cfbs/cfbs_config.py
```diff
@@ -98,6 +99,39 @@ def save(self):
     def longest_module_name(self) -> int:
         return max((len(m["name"]) for m in self["build"]))
 
+    def _handle_input(self, input):
```
vpodzime, line 102: I would actually move this into `cfbs.prompts` and drop the first commit because it would eliminate the circular imports.

vpodzime (commented):
Looks promising otherwise. Next is handling inputs on module updates and making sure everything works with dependencies (both when adding modules and updating them).

cfbs/cfbs_config.py
vpodzime: Oh, I lost track of the changes in those files! I thought `CFBSConfig()` was a config not something that actually even adds and removes modules etc. Anyway, I still think this input handling should be separated. If for nothing else then at least for better reusability.

cfbs/cfbs_config.py
larsewi: Like this?
```python
for name in {x["name"] for x in self["build"]}.difference(x["name"] for x in before):
    msg += "\n - Added module '%s'" % name
    count += 1
```
I think the current one is easier to read.

The following does not work:
```python
for module in set(self["build"]).difference(set(before)):
    msg += "\n - Added module '%s'" % module["name"]
    count += 1
```
```
Traceback (most recent call last):
...
  File "/usr/local/lib/python3.8/dist-packages/cfbs/cfbs_config.py", line 306, in add_command
    for module in set(self["build"]).difference(set(before)):
TypeError: unhashable type: 'collections.OrderedDict'
```

vpodzime: I find the (working) set difference easiest to read only I would actually assign the two sets to variables like `module_names` and `orig_module_names`.

larsewi: I agree, thanks :rocket:

cfbs/commands.py
```diff
@@ -762,3 +781,87 @@ def info_command(modules):
...
+        if key not in input:
+            user_error("Expected %s in definition: %s" % (key, pretty(input)))
+
+
+def _handle_input(input):
+    if "string" in input:
+        # TODO: implement choices
+        string = input["string"]
+        _check_keys(("question", "label"), string)
+        return prompt_user(string["question"], default=string.get("default"))
```
vpodzime: `string` is not a good variable name, AFAICT.

cfbs/commands.py
```diff
@@ -762,3 +781,87 @@ def info_command(modules):
...
+    # TODO: Implement number (should create a JSON number in augment)
+    # TODO: Implement file (should copy a file into module folder)
+
+
+def _input_command(module, input_from):
+    print("Collecting input for module '%s'" % module["name"])
+    def_json = {"variables": {}}
+
+    for input in module["input"]:
+        _check_keys((), input)
```
vpodzime: ?

cfbs/commands.py
```diff
@@ -762,3 +781,87 @@ def info_command(modules):
...
+        namespace = input["namespace"] if "namespace" in input else "cfbs"
```
vpodzime: `input.get("namespace", "cfbs")`

cfbs/commands.py
```diff
@@ -803,6 +804,46 @@ def _check_keys(keys, input):
...
+            for o in obj:
+                _check_keys(["key", "value"], o)
+                if not _validate(o["value"], value[o["key"]]):
+                    return False
+            return True
+        return False
+
+    for module in build:
+        def_json = read_json(os.path.join(module["name"], "def.json"))
+        if "input" not in module or not def_json or "variables" not in def_json:
```
vpodzime: `def_json is None`

cfbs/cfbs_config.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+from inspect import ismethoddescriptor
```
vpodzime, line 2: Looks like an unused import.

cfbs/cfbs_config.py
```diff
@@ -311,15 +311,26 @@ def add_command(
...
-            msg = msg[4:]  # Remove the '\n - ' part of the message
+
+            module = self.get_module_from_build(name)
+            if "input" in module and prompt_user(
+                self.non_interactive,
+                "The added module '%s' accepts user input. Do you want to add it now?"
+                % name,
+                YES_NO_CHOICES,
+                "no",
+            ) in ("yes", "y"):
```
vpodzime, line 326: I believe the input from user needs `input.lower()`.

cfbs/commands.py
```diff
@@ -534,24 +517,55 @@ def update_command(to_update):
...
-                    if ans.lower() in ["y", "yes"]:
+                    if ans.lower() in ("y", "yes"):
                         module["steps"] = index_info["steps"]
+                        changes_made = True
                     else:
                         print(
                             "Please make sure the old build steps work"
                             + " with the new version of the module"
                         )
+            if key == "input":
```
vpodzime, line 543: `else if`

cfbs/commands.py
```diff
@@ -534,24 +517,55 @@ def update_command(to_update):
...
+                if commit_differs:
+                    ans = prompt_user(
+                        config.non_interactive,
+                        "Module '%s' takes different input now. " % update.name
```
vpodzime, line 547: How do you know that the inputs are different?

cfbs/commands.py
```diff
@@ -534,24 +517,55 @@ def update_command(to_update):
...
+                    if ans.lower() in ("y", "yes"):
+                        module["input"] = index_info["input"]
+                        changes_made = True
+
+                        ans = prompt_user(
+                            config.non_interactive,
+                            "Do you want to add input for module '%s' right away?",
+                            choices=YES_NO_CHOICES,
+                            default="no",
+                        )
```
vpodzime, line 561: This second prompt is weird. The user just told us they wanted to use new inputs. I think the two prompts should be combined into a single one giving the user an option to use the defaults.

vpodzime (commented):
+ *instead of~f~`* in one of the commit messages

Looks generally OK to me. I don't like the idea of how this whole thing works, but that's another story.

cfbs/result.py
```diff
@@ -0,0 +1,12 @@
+class Result:
+    def __init__(self, rc, do_commit=False, commit_msg=None):
+        self.rc = rc
+        self.do_commit = do_commit
+        self.commit_msg = commit_msg
+
+    @staticmethod
+    def unpack(result) -> tuple:
```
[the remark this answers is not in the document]
larsewi: Did not know you could do default values in constructor of namedtuples. Now I know. Thanks :beers:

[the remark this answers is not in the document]
larsewi: > Changed in version 3.7: Added the defaults parameter and the _field_defaults attribute.

Apparently the defaults parameter was added in 3.7, thus we cannot use it. Any thoughts on how we could do this?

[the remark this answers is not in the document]
vpodzime: You can steal from me: https://github.com/storaged-project/blivet/blob/master/blivet/util.py#L1019

JSON.md
```diff
@@ -172,3 +172,390 @@ cfbs init && cfbs add https://github.com/cfengine/some-repo
...
+Some modules allow for users to add module input by responding to questions
+expressed under the `"input"` attribute in `cfbs.json`. User input can be added
+using the `cfbs input <module-name>` command, which stores responses in
+`./<module-name>/input.json`. These responses are translated into augments which
+will be added to `./out/masterfiles/def.json` during `cfbs build`.
+
+### Create single file example:
+The `"input"` attribute takes a list of input definitions as illustrated below.
+The value of the `"input"` attribute may be expressed as inline JSON, or as a
+file path or URL to an external JSON file.
```
vpodzime, line 187: Do we have a real reason/use-case for inputs specified in an external JSON file? If not, let's not complicate our lives by allowing this, at least not for now.

JSON.md
```diff
@@ -172,3 +172,390 @@ cfbs init && cfbs add https://github.com/cfengine/some-repo
...
+    "my_namespace:my_bundle.filename": {
+      "value": "/tmp/create-single-file.txt",
+      "comment": "Example comment."
+    }
+  }
+}
+```
+
+### Create a single file with content example:
+Modules that creates empty files is not too impressive on its own. Let us
```
vpodzime, line 300: *A module...*

JSON.md
```diff
@@ -172,3 +172,388 @@ cfbs init && cfbs add https://github.com/cfengine/some-repo
...
+    }
+  }
+}
+```
+
+From the example above we can see our beloved `filename`-variable along with a
+class generated by the autorun dependency. Studying our variable closer, we can
+see that a namespace, bundle, and a comment, were automatically added with some
+default values.  I.e.  `cfbs`, the module name canonified, and `Added by 'cfbs
+input'` respectivy.  These defaults can easily be overrided using the
```
larsewi, line 266: `s/overrided/overridden`

JSON.md
```diff
@@ -172,3 +172,388 @@ cfbs init && cfbs add https://github.com/cfengine/some-repo
...
+  },
+  "variables": {
+    "cfbs:create_single_file.filename": {
+      "value": "/tmp/create-single-file.txt",
+      "comment": "Added by 'cfbs input'"
+    }
+  }
+}
+```
+
```
larsewi, line 261: Show that a buildstep is added in `cfbs.json` after build to make this more clear
```json
"steps": [
        "copy ./create-multiple-files-with-content.cf services/autorun/create-multiple-files-with-content.cf",
        "input ./input.json def.json"
      ],
```

cfbs/commands.py
```diff
@@ -457,88 +462,158 @@ def _someone_needs_me(this) -> bool:
     return 0
 
 
+def _update_input(module):
```
larsewi, line 465: Add unit test

cfbs/cfbs_config.py
```diff
@@ -307,15 +307,31 @@ def add_command(
...
+        for name in added:
+            msg += "\n - Added module '%s'" % name
             count += 1
-        if count > 1:
-            msg = "Added %d modules\n" % count + msg
-        else:
-            msg = msg[4:]  # Remove the '\n - ' part of the message
+
+            module = self.get_module_from_build(name)
+            if (
```
larsewi, line 316: `cfbs add` should check if input needs update

cfbs/cfbs_config.py
```diff
@@ -314,3 +316,82 @@ def add_command(
...
+                    result = OrderedDict()
+                    for element in subtype:
+                        _check_keys(["type", "label", "question", "key"], element)
+                        if element["type"] != "string":
+                            user_error(
+                                "Subtype of type '%s' not supported for type list"
+                                % element["type"]
+                            )
+                        result[element["key"]] = _input_string(element)
+                    return result
```
vpodzime, line 379: This nested function somewhere inside another function is weird. I'd move it above the `_input_list()` function with one argument for the `subtype`.

cfbs/commands.py
```diff
@@ -799,3 +800,37 @@ def info_command(modules):
...
+    for module_name in args:
+        module = config.get_module_from_build(module_name)
+        if not module:
+            print("Skipping module '%s', module not found" % module_name)
+            continue
+        if "input" not in module:
+            print("Skipping module '%s', no input needed" % module_name)
+            continue
+
+        input_path = "./%s/input.json" % module_name
```
vpodzime, line 944: `os.path.join()`

vpodzime (commented):
I've only gotten to *Added command cfbs input*, will follow up with the rest of the commits later. Apart from the nitpicks below, please make sure that the docs added in the second commit still actually reflect what the commits after it introduce (and maybe move it to the end of the commits in this PR?). Thanks for working on this!

cfbs/commands.py
```diff
@@ -453,88 +458,181 @@ def _someone_needs_me(this) -> bool:
...
+def update_input_data(module, input_data):
+    """
+    Update input data from module definition
+
+    Params:
+        module     (dict): Module with updated input definition
+        input_data (dict): Input data to update
+
+    Returns:
+        bool: True if changes are made
```
vpodzime, line 469: Might as well use the _Sphinx_ syntax here (see the docstrings in *git.py*).

cfbs/commands.py
```diff
@@ -49,6 +50,10 @@
 from cfbs.module import Module
 
 
+class InputDataUpdateFailedException(Exception):
```
vpodzime, line 53: I think the `Exception` part is not needed in the name of the exception class. If you want to have something there, `Error` is usually used.

cfbs/commands.py
```diff
@@ -453,88 +458,181 @@ def _someone_needs_me(this) -> bool:
...
+    module_name = module["name"]
+    input_def = module["input"]
+
+    if len(input_def) != len(input_data):
+        log.warning(
+            "Failed to update input data for module '%s': " % module_name
+            + "Input definition has %d variables, " % len(input_def)
+            + "while current input data has %d variables." % len(input_data)
+        )
+        raise InputDataUpdateFailedException()
```
vpodzime, line 480: I think the exception should also contain the message/reason).

cfbs/commands.py
```diff
@@ -453,88 +458,181 @@ def _someone_needs_me(this) -> bool:
...
-            to_update.remove(m)
-    else:
-        # Update all modules in build if no modules are specified
-        to_update = [Module(m["name"]) for m in config["build"]]
+    # Update all modules in build if none specified
+    to_update = (
+        [Module(m) for m in to_update]
+        if to_update
+        else [Module(m["name"]) for m in build]
+    )
```
vpodzime, line 580: This part (and many changes below) doesn't seem to be related to handling inputs as part of `update`. Ideally, it should be in a separate commit, but not a big deal if it would be hard to separate.

tests/shell/all.sh
```diff
@@ -20,3 +20,7 @@ bash tests/shell/014_add_nonexistent.sh
 bash tests/shell/015_add_version.sh
 bash tests/shell/016_add_folders.sh
 bash tests/shell/017_info.sh
+bash tests/shell/018_update_input_one_variable.sh
+bash tests/shell/019_update_input_two_variables.sh
+bash tests/shell/020_update_input_list.sh
+bash tests/shell/021_update_input_list_with_keys.sh
```
vpodzime, line 26: Would be nice to have tests for the bad cases too.

cfbs/build.py
```diff
@@ -1,5 +1,7 @@
 import os
+import logging as log
```
vpodzime, line 2: *If input associated with the 'input' build step, no operation is done.* -- missing *no* in the commit message

cfbs/build.py
```diff
@@ -106,6 +144,36 @@ def _perform_build_step(module, step, max_length):
...
+            )
+            return
+        extras, original = read_json(src), read_json(dst)
+        extras = _generate_augment(module["name"], extras)
+        log.debug("Generated augment: %s", pretty(extras))
+        if not extras:
+            log.warning(
+                "Input data '%s' is incomplete: Skipping build step."
+                % os.path.basename(src)
+            )
```
vpodzime, line 168: This should be an error.

cfbs/commands.py
```diff
@@ -382,9 +382,8 @@ def _get_modules_by_url(name) -> list:
         input_path = os.path.join(".", name, "input.json")
         if os.path.isfile(input_path) and prompt_user(
             config.non_interactive,
-            "Module '%s' has input data '%s'. Do you want to remove it?"(
-                name, input_path
-            ),
+            "Module '%s' has input data '%s'. Do you want to remove it?"
+            % (name, input_path),
```
vpodzime, line 386: This fix should go to a different commit.

cfbs/cfbs_config.py
```diff
@@ -316,23 +316,26 @@ def add_command(
...
-            if (
-                "input" in module
-                and not os.path.isfile(input_path)
-                and prompt_user(
+            if "input" in module:
+                if os.path.isfile(input_path):
+                    log.warning(
+                        "There seems to already be input for module '%s'. " % name
+                        + "Note that old input may break the module. "
+                        + "Please make sure to run 'cfbs update', "
```
vpodzime, line 324: This is confusing. I'd expect `cfbs input MODULE` to be the command to run.

tests/shell/022_add_input_remove.sh
```diff
@@ -0,0 +1,14 @@
...
+mkdir -p ./tmp/
+cd ./tmp/
+touch cfbs.json && rm cfbs.json
+rm -rf .git
+rm -rf example-module
+cp ../shell/022_add_input_remove/example-cfbs.json cfbs.json
+
+cfbs --non-interactive add example-module
+cfbs --non-interactive input example-module
+cat ./example-module/input.json | grep '"response": "/tmp/testfile.txt"'
```
vpodzime, line 13: `grep` takes a file path argument ;)

vpodzime (commented):
Suggestions for the rest of the commits.

cfbs/build.py
```diff
@@ -21,6 +23,42 @@ def init_out_folder():
...
+    """
+    Generate augment from input data.
+
+    Params:
+        module_name (str): Name of module
+        input_data (list): Input data
+
+    Returns:
+        dict: Generated augment or None if input data is incomplete
+    """
```
larsewi, line 36: Change to Sphinx syntax here as well

tests/shell/021_update_input_list_with_keys.sh
```diff
@@ -0,0 +1,22 @@
...
+cat example-module/input.json | grep '"while": "Create another file?"'
+cat example-module/input.json | grep '"label": "Path"'
+cat example-module/input.json | grep '"question": "Path of file?"'
+cat example-module/input.json | grep '"default": "/tmp/test.txt"'
+cat example-module/input.json | grep '"label": "Contents"'
+cat example-module/input.json | grep '"question": "File contents?"'
+cat example-module/input.json | grep '"default": "Hello CFEngine!"'
+cat example-module/input.json | grep '"response":'
+cat example-module/input.json | grep '{ "name": "/tmp/test-1.txt", "content": "Hello CFEngine!" }'
+cat example-module/input.json | grep '{ "name": "/tmp/test-2.txt", "content": "Bye CFEngine!" }'
```
larsewi, line 22: Use grep argument here as well

tests/shell/020_update_input_list.sh
```diff
@@ -0,0 +1,20 @@
...
+
+cfbs --loglevel=debug --non-interactive update
+cat example-module/input.json | grep '"label": "Filepaths"'
+cat example-module/input.json | grep '"while": "Create another file?"'
+cat example-module/input.json | grep '"label": "Path"'
+cat example-module/input.json | grep '"question": "Path of file?"'
+cat example-module/input.json | grep '"default": "/tmp/test.txt"'
+cat example-module/input.json | grep '"response":'
+cat example-module/input.json | grep '"/tmp/create-multiple-files-1.txt"'
+cat example-module/input.json | grep '"/tmp/create-multiple-files-2.txt"'
```
larsewi, line 20: Use grep argument here as well

tests/shell/019_update_input_two_variables.sh
```diff
@@ -0,0 +1,19 @@
...
+cp ../shell/019_update_input_two_variables/example-cfbs.json cfbs.json
+
+cfbs --loglevel=debug --non-interactive update
+cat example-module/input.json | grep '"label": "Path"'
+cat example-module/input.json | grep '"question": "Path of file?"'
+cat example-module/input.json | grep '"default": "/tmp/test.txt"'
+cat example-module/input.json | grep '"label": "Contents"'
+cat example-module/input.json | grep '"question": "File contents?"'
+cat example-module/input.json | grep '"default": "Hello CFEngine!"'
+cat example-module/input.json | grep '"response": "/tmp/create-single-file-with-content.txt"'
```
larsewi, line 19: Use grep argument here as well

cfbs/utils.py
```diff
@@ -281,3 +281,7 @@ def wrapper(*args, **kwargs):
         return memo[key]
 
     return wrapper
+
+
+def canonify(string):
+    return re.sub(r"[^a-zA-Z0-9_\200-\377]", "_", string)
```
olehermanse, line 289: Should probably reference where this is from (in commit message).

tests/shell/025_add_input_remove.sh
```diff
@@ -0,0 +1,14 @@
...
+cd ./tmp/
+touch cfbs.json && rm cfbs.json
+rm -rf .git
+rm -rf example-module
+cp ../shell/025_add_input_remove/example-cfbs.json cfbs.json
+
+cfbs --non-interactive add example-module
+cfbs --non-interactive input example-module
+grep '"response": "/tmp/testfile.txt"' ./example-module/input.json
+cfbs --non-interactive remove example-module
```
olehermanse, line 14: Should probably check here what remove did.

olehermanse (approved):
Looks good to me, thanks for working on this!

PR #126, opened 2022-07-26
ENT-9024: Add custom indexes to each module config as they are added

craigcomstock:
Ticket: ENT-9024
Changelog: title

### DIFF
--- a/cfbs/index.py
+++ b/cfbs/index.py
@@ -98,6 +98,15 @@ class Index:
             self._expand_index()
         return self._data
 
+    @property
+    def custom_index(self) -> str:
+        # Index can be initialized with a dict or OrderedDict instead of a url string
+        # in which case there would not be an index url to check against the default
+        if type(self._unexpanded) is str and self._unexpanded != _DEFAULT_INDEX:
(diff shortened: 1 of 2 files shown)


cfbs/index.py
```diff
@@ -98,6 +101,10 @@ def data(self) -> dict:
             self._expand_index()
         return self._data
 
+    @property
+    def url(self) -> str:
+        return self._unexpanded
+
```
olehermanse, line 107: The reason why this attribute is not called URL is because it can also be just a dict with the data. See the logic in `_expand_index()`.

So, I don't agree with this change, you can either change the internals to have 2 different variables: unexpanded `url` and `data` (both of which can be None), or you can change the `url` method to return `None` if `self._unexpanded` is dict:

```suggestion
    @property
    def url(self) -> str:
        if type(self._unexpanded) is not str:
            return None
        return self._unexpanded

```

craigcomstock: Yeah, makes sense. Made a change.

olehermanse (approved):

PR #127, opened 2022-09-23
Use "main" as the initial branch for 'git init'

vpodzime:
Seems to be what the world is moving to from calling it "master" and we cannot rely on the default.

Ticket: ENT-9150
Changelog: None


larsewi (approved):
ACK

olehermanse (approved):

PR #128, opened 2022-09-29
ENT-9351: Added ability to download modules added by URL

craigcomstock:
Key off the absence of a "version" property in the module reference in cfbs.json to know to git clone instead of trying to download and checksum the archive.

Ticket: ENT-9351
Changelog: title


nickanderson (approved):

nickanderson (approved):
I confirmed this addresses my issue. With the patch I am able to take a fresh hub, point Build in MP at a repo with a previously generated cfbs.json that includes modules that were added directly from module repos instead of the build index and the policy is built and deployed as expected.

PR #130, opened 2022-09-29
ENT-9351: Added ability to download modules added by URL

craigcomstock:
Key off the absence of a "version" property in the module reference in cfbs.json to know to git clone instead of trying to download and checksum the archive.

Ticket: ENT-9351
Changelog: title

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -623,7 +623,10 @@ def _download_dependencies(
         if not os.path.exists(module_dir):
             if url.endswith(SUPPORTED_ARCHIVES):
                 fetch_archive(url, commit)
-            elif "index" in module or ignore_versions:
+            # a couple of cases where there will not be an archive available:
+            # - using an alternate index (index property in module data)
+            # - added by URL instead of name (no version property in module data)
+            elif "index" in module or not "version" in module or ignore_versions:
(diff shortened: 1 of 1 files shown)


nickanderson (approved):
I mentioned it in the previous PR for this https://github.com/cfengine/cfbs/pull/128

But i confirmed that this does fix my issue.

vpodzime (approved):
ACK

cfbs/commands.py
```diff
@@ -623,7 +623,10 @@ def _download_dependencies(
         if not os.path.exists(module_dir):
             if url.endswith(SUPPORTED_ARCHIVES):
                 fetch_archive(url, commit)
-            elif "index" in module or ignore_versions:
+            # a couple of cases where there will not be an archive available:
+            # - using an alternate index (index property in module data)
+            # - added by URL instead of name (no version property in module data)
+            elif "index" in module or not "version" in module or ignore_versions:
```
olehermanse, line 629: The module added by URL should have a `"url"` key.

I think this is the more appropriate way to fix it:

```suggestion
            elif "index" in module or if "url" in module or ignore_versions:
```

olehermanse:
@craigcomstock I will put your commit into this PR with a test as well:

https://github.com/cfengine/cfbs/pull/131

PR #131, opened 2022-09-30
ENT-9351: Test and fix for cfbs build on modules added by URL

olehermanse:


olehermanse:
First commit adds a test which reproduces the issue, second commit (thanks @craigcomstock) fixes the issue, as can be seen from CI status checkmark.

vpodzime (approved):
ACK

PR #132, opened 2022-10-11
ENT-9366: Fix updating renamed modules that alias the old name

craigcomstock:
Ticket: ENT-9366


craigcomstock:
Sadly, this PR doesn't quite do everything I think it should. It does handle getting the new data from a module that used to be "real" and is now an alias. But it doesn't change the module name to the new name as `cfbs add` would if given the alias. So maybe a TODO there.

craigcomstock:
close this for now, not actively working on fixing.

PR #136, opened 2022-10-26
CFE-4088: Made policy_files build step not implicitly add services/cfbs

olehermanse:
Also improved the documentation, commits:

- Made policy_files build step not implicitly add services/cfbs
- Updated tests according to latest changes in policy_files build step
- Added success message after all shell tests have run
- README.md: Added documentation for the new build steps; bundles, policy_files, input
- Moved explanation of build steps into JSON.md


README.md
```diff
@@ -105,6 +105,48 @@ cfbs clean
 cfbs input create-single-file
 ```
 
+This command prompts the user for module input to configure what the module should do.
+
+Only works for modules which accept input, indicated by the `"input"` key.
+This `input` key contains the specification for what input the module accepts, if any.
+
+User's responses are stored in `<module-name>/input.json`.
```
nickanderson, line 113: I still really do not like this. It makes it difficult to tell what's cfbs managed and what's my own stuff in the cfbs project. I thin this would be much better off under a common top level directory like `cfbs-module-input/<module-name>/input.json`

README.md
```diff
@@ -257,3 +299,20 @@ Destination is relative to the output policy set (`out/masterfiles`).
 - Enable `services_autorun_bundles` class in `def.json`.
 - Merge any `def.json` recursively into `out/masterfiles/def.json`.
 - Copy any other files with their existing directory structure to destination.
+
+#### `bundles <bundles ...>`
+- Ensure bundles are evaluated by adding them to the bundle sequence, using `def.json`.
+- Only manipulates the bundle sequence, to ensure policy files are copied and parsed, use other build steps, for example `copy` and `policy_files`.
```
nickanderson: Specifically, this adds to `default:def.control_common_bundlesequence_end` right? That's an MPFism, so I think it makes sense to call it out specifically.

README.md
```diff
@@ -257,3 +299,20 @@ Destination is relative to the output policy set (`out/masterfiles`).
...
+
+#### `policy_files <paths ...>`
+- Add policy (`.cf`) files to `inputs` key in `def.json`, ensuring they are parsed.
```
nickanderson: So, the inputs key, is also an MPFism. It ultimately populates `default:def.augments_inputs` which then must be included in inputs by the actual policy.

Initially when augments parsing moved to C from policy the inputs key would inject directly to inputs, so that applied for any policy entry that loaded the augments file. We saw some issues with duplicate definition of bundle and we backed off since direct injection to inputs provided no way to vary inputs per entry (e.g. different policy files added to inputs for `promises.cf` vs `update.cf`).

So, currently `promises.cf` uses `default:def.augments_inputs` to extend inputs but update.cf looks for `default:def.update_inputs`, which must be specified in augments via `vars` or `variables`.

Anyway, I think the wording here is a bit strong given how it actually works and how much of it is specific to MPF.

JSON.md
```diff
@@ -16,6 +16,85 @@ All these 3 share 1 standard format, commonly called `cfbs.json`.
...
+
+```json
+{
+  "name": "./policy.cf",
+  "description": "Local policy file added using cfbs command line",
+  "tags": ["local"],
+  "dependencies": ["autorun"],
+  "steps": ["copy ./policy.cf services/autorun/policy.cf"],
+  "added_by": "cfbs add"
+}
```
larsewi, line 34: Local policy files no longer depend on autorun. We should update this example to adhere to the modern world of cfbs.
```suggestion
{
  "name": "./policy.cf",
  "description": "Local policy file added using cfbs command line",
  "tags": ["local"],
  "added_by": "cfbs add",
  "steps": [
    "copy ./policy.cf services/cfbs/policy.cf",
    "policy_files ./policy.cf",
    "bundles my_bundle"
  ]
}
```

README.md
```diff
@@ -257,3 +299,20 @@ Destination is relative to the output policy set (`out/masterfiles`).
...
+  - Only edits `def.json`, does not copy files. Should be used after a `copy` or `directory` build step.
+  - Does not add any bundles to the bundle sequence, to ensure a bundle is evaluated, use the `bundles` build step or the autorun mechanism.
+- All paths are relative to `out/masterfiles`.
+- If any of the paths are directories (end with `/`), the folder(s) are searched and all `.cf` files are added.
```
nickanderson: searched recursively?

nickanderson (changes requested):
Helpful docs here, just some comments about details that could result in users being mislead about behavior.

README.md
```diff
@@ -178,82 +220,7 @@ from pydantic import (
...
-#### `append <source> <destination>`
-- Append the source file to the end of destination file.
+## The cfbs.json format
 
-#### `directory <source> <destination>`
-- Copy any .cf policy files recursively and add their paths to `def.json`'s `inputs`.
-- Enable `services_autorun_bundles` class in `def.json`.
-- Merge any `def.json` recursively into `out/masterfiles/def.json`.
-- Copy any other files with their existing directory structure to destination.
+More advanced users and module authors may need to understand, write, or edit `cfbs.json` files.
```
larsewi, line 225:
```suggestion
More advanced users and module authors may need to understand how to write or edit `cfbs.json` files.
```

larsewi (approved):
Thanks!

README.md
larsewi: If you use a directory as argument instead of a policy file; it will search through the directory and its sub-directories for any files that with the `.cf` file extension and add them to inputs.

README.md
olehermanse: This is the default, we can discuss adding the possibility to customize it: https://tracker.mender.io/browse/CFE-4091

README.md
olehermanse: I can try to make this a bit more clear, but in general MPF will be assumed / required for some features of CFEngine and cfbs.

JSON.md
```diff
@@ -16,6 +16,85 @@ All these 3 share 1 standard format, commonly called `cfbs.json`.
...
+- Merge any `def.json` recursively into `out/masterfiles/def.json`.
+- Copy any other files with their existing directory structure to destination.
+
+#### `bundles <bundles ...>`
+- Ensure bundles are evaluated by adding them to the bundle sequence, using `def.json`.
+- Only manipulates the bundle sequence, to ensure policy files are copied and parsed, use other build steps, for example `copy` and `policy_files`.
+
+#### `policy_files <paths ...>`
+- Add policy (`.cf`) files to `inputs` key in `def.json`, ensuring they are parsed.
+  - Only edits `def.json`, does not copy files. Should be used after a `copy` or `directory` build step.
```
olehermanse, line 87:
```suggestion
  - Note that this relies on using the default policy set from the CFEngine team, the Masterfiles Policy Framework, commonly added as the first module (`masterfiles`).
  - Only edits `def.json`, does not copy files. Should be used after a `copy` or `directory` build step.
```

JSON.md
```diff
@@ -16,6 +16,85 @@ All these 3 share 1 standard format, commonly called `cfbs.json`.
...
+- Typically used before copying files to the output policy set with the `copy` step. 
+
+#### `directory <source> <destination>`
+- Copy any .cf policy files recursively and add their paths to `def.json`'s `inputs`.
+- Enable `services_autorun_bundles` class in `def.json`.
+- Merge any `def.json` recursively into `out/masterfiles/def.json`.
+- Copy any other files with their existing directory structure to destination.
+
+#### `bundles <bundles ...>`
+- Ensure bundles are evaluated by adding them to the bundle sequence, using `def.json`.
```
olehermanse, line 82:
```suggestion
- Ensure bundles are evaluated by adding them to the bundle sequence, using `def.json`.
  - Note that this relies on using the default policy set from the CFEngine team, the Masterfiles Policy Framework, commonly added as the first module (`masterfiles`). 
    Specifically, this build step adds the bundles to the variable `default:def.control_common_bundlesequence_end`, which the MPF looks for.
```

README.md
olehermanse: No, the intention here was 3 different activities: understand the JSONs, write them, and edit them.

README.md
olehermanse: Fixed.

README.md
olehermanse: Added clarification.

JSON.md
```diff
@@ -16,6 +16,91 @@ All these 3 share 1 standard format, commonly called `cfbs.json`.
...
+    Specifically, this build step adds the bundles to the variable `default:def.control_common_bundlesequence_end`, which the MPF looks for.
+- Only manipulates the bundle sequence, to ensure policy files are copied and parsed, use other build steps, for example `copy` and `policy_files`.
+
+#### `policy_files <paths ...>`
+- Add policy (`.cf`) files to `inputs` key in `def.json`, ensuring they are parsed.
+  - Note that this relies on using the default policy set from the CFEngine team, the Masterfiles Policy Framework, commonly added as the first module (`masterfiles`).
+  - Only edits `def.json`, does not copy files. Should be used after a `copy` or `directory` build step.
+  - Does not add any bundles to the bundle sequence, to ensure a bundle is evaluated, use the `bundles` build step or the autorun mechanism.
+- All paths are relative to `out/masterfiles`.
+- If any of the paths are directories (end with `/`), the folder(s) are searched and all `.cf` files are added.
```
olehermanse, line 96:
```suggestion
- If any of the paths are directories (end with `/`), the folder(s) are recursively searched and all `.cf` files are added.
```

JSON.md
```diff
@@ -16,6 +16,91 @@ All these 3 share 1 standard format, commonly called `cfbs.json`.
...
+  "added_by": "cfbs add",
+  "steps": [
+    "copy ./policy.cf services/cfbs/policy.cf",
+    "policy_files ./policy.cf",
+    "bundles my_bundle"
+  ]
+}
+```
+
+In the example above, the `policy.cf` file is copied to `out/masterfiles/services/autorun/policy.cf`.
```
olehermanse, line 40:
```suggestion
The 3 build steps above achieve 3 distinct things:
1. The policy file `policy.cf` is included in the policy set. It will be deployed to `/var/cfengine/masterfiles/services/cfbs/policy.cf` on the hub.
2. The path to `policy.cf` is added to `"inputs"` making cf-agent and other binaries aware of it and parse it.
3. The bundle `my_bundle` within `policy.cf` is added to the bundle sequence, making cf-agent find the correct bundle to run and, in the end, enforcing the promises within (commonly called the entry point of a policy file).

```

JSON.md
```diff
@@ -16,6 +16,91 @@ All these 3 share 1 standard format, commonly called `cfbs.json`.
...
+
+```json
+{
+  "name": "./policy.cf",
+  "description": "Local policy file added using cfbs command line",
+  "tags": ["local"],
+  "added_by": "cfbs add",
+  "steps": [
+    "copy ./policy.cf services/cfbs/policy.cf",
+    "policy_files ./policy.cf",
```
olehermanse, line 34:
```suggestion
    "policy_files ./services/cfbs/policy.cf",
```

nickanderson (approved):
:people_hugging:

PR #139, opened 2022-10-28
CFE-4070: Added commands `cfbs get-input` and `cfbs set-input`

larsewi:


cf-bottom:
Thanks for submitting a pull request! Maybe @olehermanse can review this?

cfbs/commands.py
```diff
@@ -1000,3 +1001,103 @@ def input_command(args, input_from="cfbs input"):
...
+
+    spec = module.get("input")
+    if spec is None:
+        log.error("Module '%s' does not accept input" % name)
+        return 1
+    log.debug("Input spec for module '%s': %s" % (name, pretty(spec)))
+
+    if src_path is None:
+        try:
+            data = json.load(stdin)
```
olehermanse, line 1021: Use `OrderedDict` explicitly so python 3.5 also has predictable dictionaries:

```suggestion
            data = json.load(stdin, object_pairs_hook=OrderedDict)
```

cfbs/commands.py
```diff
@@ -7,6 +7,7 @@
 import copy
 import logging as log
 import json
```
olehermanse, line 9:
```suggestion
import json
from collections import OrderedDict
```

olehermanse (changes requested):
Python 3.5 not working as 3.6 is the first one were dictionaries are always oredered.

tests/shell/030_get_set_input.sh
```diff
@@ -0,0 +1,99 @@
...
+  {
+    "type": "string",
+    "variable": "bogus",
+    "label": "Filename",
+    "question": "What file should this module create?",
+    "response": "/tmp/test-2.txt"
+  }
+]' > igors-input.json
+
+! cfbs set-input create-single-file igors-input.json
```
olehermanse, line 97: Could you add another test like this where one of the keys is changed, like `type` is changed to something else?

larsewi: Sure thing! That a key is changed instead of value?

cfbs/commands.py
```diff
@@ -1000,3 +1001,74 @@ def input_command(args, input_from="cfbs input"):
...
+
+    spec = module.get("input")
+    if spec is None:
+        log.error("Module '%s' does not accept input" % name)
+        return 1
+    log.debug("Input spec for module '%s': %s" % (name, pretty(spec)))
+
+    if src_path is None:
+        try:
+            data = json.load(stdin)
```
vpodzime: If the user runs the command `cfbs set-input some-module`, they will get a seemingly stuck/hanging tool. We should either add some prompt or change the stdin-based input to be based on an extra option or a special value of the argument (`-` is typically used for this).

cfbs/commands.py
```diff
@@ -1000,3 +1001,74 @@ def input_command(args, input_from="cfbs input"):
...
+
+    assert data is not None
+    log.debug("Input data for module '%s': %s" % (name, pretty(data)))
+
+    def _compare_dict(a, b, ignore={}):
+        assert isinstance(a, dict) and isinstance(b, dict)
+        if set(a.keys()) != set(b.keys()) - ignore:
+            return False
+        """Avoid code duplication by converting the values of the two dicts
+        into two lists in the same order and compare the lists instead"""
```
vpodzime, line 1040: Please use comments for what is commenting the code. Documentation strings are for documentation.

cfbs/main.py
```diff
@@ -119,6 +119,11 @@ def main() -> int:
         return commands.update_command(args.args)
     if args.command == "input":
         return commands.input_command(args.args)
+    if args.command == "set-input":
+        if len(args.args) < 1:
+            user_error("Missing argument <module-name>")
+        module, file = args.args[0], args.args[1] if len(args.args) >= 2 else None
```
vpodzime, line 125: Please use parentheses to make this easier to read.

cfbs/commands.py
```diff
@@ -1000,3 +1001,74 @@ def input_command(args, input_from="cfbs input"):
...
+        keys = a.keys()
+        return _compare_list([a[key] for key in keys], [b[key] for key in keys])
+
+    def _compare_list(a, b):
```
vpodzime, line 1044: So this requires that things are in the very same order. I'm not sure how much strict JSON is regarding ordering of object's children, but it feels wrong to me to require the inputs to be in the very same order.

cfbs/commands.py
```diff
@@ -1072,3 +1072,32 @@ def _compare_list(a, b):
...
+        data = module["input"]
+    else:
+        log.debug("Loaded input from '%s'" % src_path)
+
+    if dst_path is None:
+        log.debug("Writing json to stdout")
+        print(pretty(data))
+    else:
+        log.debug("Writing json to '%s'" % dst_path)
+        write_json(dst_path, data)
```
vpodzime, line 1102: I would expect `get-input module some-file` to extract the input from `some-file`. Again, I'd prefer an extra option for the output.

vpodzime (commented):
Looks OK to me otherwise.

cfbs/commands.py
larsewi: Objects does not have to be in order. I added a test to prove that. But lists do, and they should

cfbs/main.py
```diff
@@ -119,6 +119,13 @@ def main() -> int:
         return commands.update_command(args.args)
     if args.command == "input":
         return commands.input_command(args.args)
+    if args.command in ("set-input", "get-input"):
+        if len(args.args) < 1:
+            user_error("Missing argument <module-name>")
+        module, file = args.args[0], (args.args[1] if len(args.args) >= 2 else None)
+        if args.command == "set-input":
+            return commands.set_input_command(module, file)
+        return commands.get_input_command(module, file)
```
olehermanse, line 128: As @vpodzime pointed out, you can consider adding a requirement to specify filename or `-` for stdin here;

```
cfbs get-input <module-name> <filename or - for stdin>
cfbs set-input <module-name> <filename or - for stdin>
```

Then you can always require `len(args.args) == 2`, and print out a helpful error message:

```
error: Filename required as last argument (or - for stdin)
```

I don't think we should make it more complicated than this by using `--`-options or prompts.

tests/shell/030_get_set_input.sh
olehermanse: yep, thanks.

olehermanse (approved):
LGTM. There are some comments above which you can read / considers, but no blockers from my side.

cfbs/commands.py
vpodzime: Aren't the inputs in a list?

cfbs/commands.py
larsewi: Made the second argument required, and added support for the special value `-`, which is `stdin`/`stdout` for get-input/set-input respectively

cfbs/commands.py
larsewi: Yep! Imagine the following variables being swapped
```
$ cfbs input create-single-file-with-content
--snip--
What file do you want to create? /tmp/testfile
What content should this file have? Hello CFEngine!
```
i.e.
```
$ cfbs input create-single-file-with-content
--snip--
What content should this file have? Hello CFEngine!
What file do you want to create? /tmp/testfile
```
it would not make much sense for the user.

cfbs/main.py
```diff
@@ -119,6 +120,34 @@ def main() -> int:
         return commands.update_command(args.args)
     if args.command == "input":
         return commands.input_command(args.args)
+    if args.command == "set-input":
+        if len(args.args) != 2:
+            log.error("Filename required as argument (or - for stdin)")
```
olehermanse, line 125:
```suggestion
            if len(args.args) == 1:
                log.error("Filename required as argument (or - for stdin)")
            if len(args.args) == 0:
                log.error("Module and filename (or - for stdin) must be specified")
            if len(args.args) > 2:
                log.error("Too many arguments, set-input only accepts 2 (module and filename)")
```

cfbs/main.py
```diff
@@ -119,6 +120,34 @@ def main() -> int:
...
+        except OSError as e:
+            log.error("Can't open '%s': %s" % (args.args[1], e))
+            return 1
+        try:
+            return commands.set_input_command(module, infile)
+        finally:
+            infile.close()
+    if args.command == "get-input":
+        if len(args.args) != 2:
+            log.error("Filename required as argument (or - for stdout)")
```
olehermanse, line 139: Same here.

olehermanse (changes requested):
Almost ;)

cfbs/commands.py
vpodzime: It wouldn't. But when the data comes from a form in MP or generally from some JSON, ordering should not matter.

olehermanse: The order of the list defines the order of the prompts and the order of the UI elements in MP. It matters. MP should be working on the list it got from get-input, and should not reorder the elements before running set-input.

olehermanse: We could enable module authors to more easily change input (reorder it) and have it work automatically with `cfbs update`, but that's not this command.

olehermanse (approved):
LGTM now

cfbs/commands.py
vpodzime: > should not reorder the elements before running set-input
I don't see a good reason for this to be disallowed/unsupported.

olehermanse: The order here communicates how it should be shown to the user and when it comes back, how it was shown to the user.

For example, things coming back in another order could indicate that MP is not showing things correctly, or not handling the input received from the user correctly.

It is important to record what the user saw and what they responded, including the order of things. @larsewi showed one example. Similar situations can apply to the GUI. If labels are bad or confusing, or users don't read them, or module authors made mistakes when filling out the labels or variable names, having an accurate record of what the user saw and entered is valuable.

I would take it one step further and say we should also record whether the user saw the prompt (used the CLI) or whether they saw the label (used MP). However this is a bit more difficult, and we are not doing that currently.

vpodzime: I guess I'm thinking about this differently -- the input, to me, is just data. So I would expect the `get-input` command to only take answers, basically as `input_id: value` pairs

PR #142, opened 2022-11-15
Fix type error bug in `cfbs set-input` command

larsewi:
Fixed bug where operands of type `dict` and `set` where used with the `-` operator.

Ticket: None
Changelog: None

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -1023,7 +1023,7 @@ def set_input_command(name, infile):
         return 1
     log.debug("Input data for module '%s': %s" % (name, pretty(data)))
 
-    def _compare_dict(a, b, ignore={}):
+    def _compare_dict(a, b, ignore=set()):
         assert isinstance(a, dict) and isinstance(b, dict)
         if set(a.keys()) != set(b.keys()) - ignore:
             return False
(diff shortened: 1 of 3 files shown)


olehermanse (approved):
LGTM, thanks! :)

cfbs/commands.py
```diff
@@ -1023,7 +1023,7 @@ def set_input_command(name, infile):
         return 1
     log.debug("Input data for module '%s': %s" % (name, pretty(data)))
 
-    def _compare_dict(a, b, ignore={}):
+    def _compare_dict(a, b, ignore=set()):
```
vpodzime, line 1026: This was and is wrong. The dict/set used as the default value is only created once and then its modifications are kept between the function calls.

cfbs/commands.py
```diff
@@ -1051,7 +1051,7 @@ def _compare_list(a, b):
         if (
             not isinstance(a, dict)
             or not isinstance(b, dict)
-            or not _compare_dict(a, b, ignore={"response"})
+            or not _compare_dict(a, b, ignore=set({"response"}))
```
vpodzime, line 1054: So this is creating a set from a dict. Would make more sense to use a tuple or a list here.

vpodzime (commented):
@larsewi Please fix at least the default arg value.

cfbs/commands.py
olehermanse: @vpodzime no, curly braces in python can mean both set or dict. This change was not needed. When there is just one or more values, separated by commas, no colons to indicate key-value pairs, it's a set.

larsewi:
```python
>>> type({"response"}).__name__
'set'
```

PR #143, opened 2022-11-21
CFE-4099: Fixed bug where order of keys breaks set-input

larsewi:
Fix bug where order of keys breaks set-input

Fixed bug where changing the order of keys when using `cfbs set-input`
causes it to error with the following error message `ERROR: Input data
for module 'delete-files' does not conform with input definition`

Ticket: CFE-4099
Changelog: None

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -1023,9 +1023,10 @@ def set_input_command(name, infile):
         return 1
     log.debug("Input data for module '%s': %s" % (name, pretty(data)))
 
-    def _compare_dict(a, b, ignore=set()):
+    def _compare_dict(a, b, ignore=None):
+        assert ignore is None or isinstance(ignore, set)
         assert isinstance(a, dict) and isinstance(b, dict)
-        if set(a.keys()) != set(b.keys()) - ignore:
+        if set(a.keys()) != set(b.keys()) - (set() if ignore is None else ignore):
             return False
         # Avoid code duplication by converting the values of the two dicts
         # into two lists in the same order and compare the lists instead
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -1039,12 +1039,18 @@ def set_input_command(name, infile):
         for x, y in zip(a, b):
             if type(x) != type(y):
                 return False
-            if isinstance(x, dict) and not _compare_dict(x, y):
-                return False
-            if isinstance(x, list) and not _compare_list(x, y):
-                return False
-            if x != y:
-                return False
+            if isinstance(x, dict):
+                if not _compare_dict(x, y):
+                    return False
+            elif isinstance(x, list):
+                if not _compare_list(x, y):
+                    return False
+            else:
+                assert isinstance(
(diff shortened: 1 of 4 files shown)


olehermanse (approved):
LGTM, thanks :)

cfbs/commands.py
```diff
@@ -1039,12 +1039,13 @@ def _compare_list(a, b):
...
-            if isinstance(x, dict) and not _compare_dict(x, y):
-                return False
-            if isinstance(x, list) and not _compare_list(x, y):
-                return False
-            if x != y:
-                return False
+            if isinstance(x, dict) and _compare_dict(x, y):
+                continue
+            if isinstance(x, list) and _compare_list(x, y):
+                continue
```
vpodzime, line 1045: Are you sure this is the right thing to do? Please extend the commit message with an explanation of how exactly this change fixes the issue.

vpodzime (commented):
So this resolves https://github.com/cfengine/cfbs/pull/139#discussion_r1009534663, thanks. Can you also please fix https://github.com/cfengine/cfbs/pull/142/files#r1026193657 here?

cfbs/commands.py
olehermanse: Talked to Lars in person. Although the current approach works, it's probably safer (and more correct) to just do a choice between the 3 options (dict, list, other) and then do the correct comparison inside that block. The current approach with `and` and this kind of "fallthrough" to the logic below, is confusing.

Specifically, it is a bit weird to think about what happens when it is a dict, but `_compare_dict()` returns False, causing us to go down to comparing them again.

olehermanse (approved):
Thanks!

cfbs/commands.py
```diff
@@ -1023,9 +1023,10 @@ def set_input_command(name, infile):
         return 1
     log.debug("Input data for module '%s': %s" % (name, pretty(data)))
 
-    def _compare_dict(a, b, ignore=set()):
+    def _compare_dict(a, b, ignore=None):
+        assert ignore is None or isinstance(ignore, set)
         assert isinstance(a, dict) and isinstance(b, dict)
-        if set(a.keys()) != set(b.keys()) - ignore:
+        if set(a.keys()) != set(b.keys()) - (set() if ignore is None else ignore):
```
vpodzime, line 1029: The common trick is to do `ignore = ignore or set()` (or `ignore = ignore if ignore is not None else set()`) at the beginning of the function.

tests/shell/032_get_set_input_unordered.sh
```diff
@@ -0,0 +1,42 @@
...
+        "label": "Why",
+        "question": "Why should this file be deleted?",
+        "type": "string"
+      }
+    ],
+    "type": "list",
+    "variable": "files",
+    "while": "Specify another file you want deleted on your hosts?"
+  }
+]' | cfbs --log=debug set-input delete-files -
```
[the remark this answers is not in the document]
olehermanse: To me it's nicer / easier the way he wrote it

PR #144, opened 2022-11-22
CFE-4081: Fixed commiting of changes made by set-input

olehermanse:
More work remains to fix committing in `cfbs add`.

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -41,6 +41,7 @@ from cfbs.internal_file_management import (
 from cfbs.index import _VERSION_INDEX, Index
 from cfbs.git import (
     is_git_repo,
+    git_commit,
     git_get_config,
     git_set_config,
     git_init,
@@ -1067,9 +1068,21 @@ def set_input_command(name, infile):
             return 1
 
     path = os.path.join(name, "input.json")
-    log.debug("Writing json to file '%s'" % path)
-    write_json(path, data)
 
+    log.debug("Comparing with data already in file '%s'" % path)
+    old_data = read_json(path)
+    if old_data == data:
+        log.debug("Input data for '%s' unchanged, nothing to write / commit" % name)
+        return 0
+
+    log.debug("Input data for '%s' changed, writing json to file '%s'" % (name, path))
+    write_json(path, data)
+    if config.get("git", False):
+        git_commit(
+            "Set the input for module '%s'" % name,
+            edit_commit_msg=False,
+            scope=[path],
+        )
     return 0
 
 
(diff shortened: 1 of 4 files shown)


larsewi (approved):
ACK

tests/shell/032_set_input_unordered.sh
```diff
@@ -40,3 +40,12 @@ echo '[
...
 ]' | cfbs --log=debug set-input delete-files -
+
+# Error if the file has never been added:
+git ls-files --error-unmatch delete-files/input.json
+
+# Error if there are staged (added, not yet commited)
+git diff --exit-code --staged
+
+# Error if there are uncommited changes (to tracked files):
+git diff --exit-code
```
vpodzime, line 51: Looks like these checks should be in a nicely-named function in some shared bash source.

olehermanse: yep, good idea

cfbs/commands.py
```diff
@@ -1067,9 +1068,21 @@ def _compare_list(a, b):
...
 
     path = os.path.join(name, "input.json")
-    log.debug("Writing json to file '%s'" % path)
-    write_json(path, data)
 
+    log.debug("Comparing with data already in file '%s'" % path)
+    old_data = read_json(path)
+    if old_data == data:
+        log.debug("Input data for '%s' unchanged, nothing to write / commit" % name)
+        return 0
```
vpodzime, line 1076: This should, ideally, be done before the more expensive checks/comparison above.

cfbs/commands.py
```diff
@@ -1067,9 +1068,21 @@ def _compare_list(a, b):
...
+
+    log.debug("Input data for '%s' changed, writing json to file '%s'" % (name, path))
+    write_json(path, data)
+    if config.get("git", False):
+        git_commit(
+            "Set the input for module '%s'" % name,
+            edit_commit_msg=False,
```
vpodzime, line 1083: No interactivity here and prompting the user if they want to modify the commit message?

olehermanse: Right, I am submitting some changes which should explain / clear this up.

cfbs/commands.py
olehermanse: I was thinking it makes sense to always do the expensive checks (validation), even if the (potentially bad) data is already there.

cfbs/commands.py
olehermanse: WIP: https://github.com/cfengine/cfbs/pull/145

PR #148, opened 2022-11-28
Fixed commit message in the `cfbs add` command

larsewi:
Fixed issue where commit message when adding a single module with input
does not separate the long- & short description with an additional
newline.

Commit message when adding a single module with input now looks like
this:
```
Added module 'delete-files'

 - Added input for module 'delete-files'
```

Ticket: None
Changelog: None


vpodzime (approved):
I wonder when exactly this change will introduce extra blank lines in commit messages when we change something in the code above.

PR #149, opened 2022-12-09
Removed unnecessary trailing slashes

olehermanse:


olehermanse:
Merging because the failures are unrelated (this PR actually fixes failures, does not introduce any).

PR #150, opened 2022-12-09
Changed from ubuntu-latest to ubuntu 20

olehermanse:
ubuntu-latest gives you the russian roulette feature of sometimes picking 22 and sometimes 20. If you get 22, python 3.6 and 3.5 will fail.


nickanderson (approved):

oleorhagen (commented):
Great feature!

PR #153, opened 2022-12-20
ENT-9710: Switched to pythonic mkdir -p

craigcomstock:
We don't want to use sh() for cfbs commands which
are used in the Mission Portal Build application.

Ticket: ENT-9710


larsewi (approved):
ACK

PR #158, opened 2023-04-25
Fixed input build step for modules which use subdirectory

olehermanse:
Using subdirectory key from module here is incorrect, the
input.json file is just in the folder with the same
name as the module.

This change makes it work as described in the documentation:

https://github.com/cfengine/cfbs/blob/master/JSON.md


craigcomstock (approved):
yeah, I tried the change and it worked!

PR #159, opened 2023-09-20
CFE-4527: Change delete step and init --masterfiles behavior

craigcomstock:
- Changed delete step behavior to allow for non-existent paths
- Changed init --masterfiles implementation to add with commit instead of modify index version of masterfiles module

### DIFF
--- a/cfbs/utils.py
+++ b/cfbs/utils.py
@@ -61,8 +61,8 @@ def rm(path: str, missing_ok=False):
         return False
     if os.path.isdir(path):
         rmtree(path)
-    if os.path.isfile(path):
-        os.remove(path)
+    else: # Assume path is a file
+        os.remove(path) # Will raise exception if missing
     return True
 
 
--- a/cfbs/utils.py
+++ b/cfbs/utils.py
@@ -5,6 +5,7 @@ import json
 import copy
 import subprocess
 import hashlib
+import logging as log
 import urllib
 import urllib.request  # needed on some platforms
 from collections import OrderedDict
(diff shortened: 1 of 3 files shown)


cfbs/utils.py
```diff
@@ -56,6 +57,8 @@ def touch(path: str):
 def rm(path: str, missing_ok=False):
     if not missing_ok:
         assert os.path.exists(path)
+    if missing_ok and not os.path.exists(path):
+        log.warning("delete: path '%s' does not exist, skipping" % path)
```
larsewi, line 61: I would not put this warning here, because we might use it places where we expect the operation to fail. However you can change the `rm()` function to return `True` or `False` based on the success of the operation. Then you can make decisions based on this condition elsewhere.

larsewi (changes requested):
Nice work 🚀

cfbs/utils.py
craigcomstock: Sure, that makes sense. I can refactor back to the "delete" step.

larsewi (approved):
ACK

cfbs/utils.py
```diff
@@ -56,10 +57,13 @@ def touch(path: str):
 def rm(path: str, missing_ok=False):
     if not missing_ok:
         assert os.path.exists(path)
+    if missing_ok and not os.path.exists(path):
+        return False
     if os.path.isdir(path):
         rmtree(path)
     if os.path.isfile(path):
         os.remove(path)
```
olehermanse, line 65:
```suggestion
    else: # Assume path is a file
        os.remove(path) # Will raise exception if missing
```

olehermanse (approved):

PR #163, opened 2023-11-12
Removed tests from the installed python package

oleorhagen:
The current build installs the tests files into site-packages in the `tests` folder, which is probably not what we want.


cf-bottom:

olehermanse (approved):

larsewi:
Nice 🚀

PR #164, opened 2023-11-29
ENT-10944: cfbs build/download: Added validation and error messages

olehermanse:
Added many helpful error messages to tell the user what
is wrong with their cfbs.json (why they cannot run cfbs
build or cfbs download).

The config must have a build field, with a list of modules,
each module must have a name, a non-empty list of steps,
each step must be a non-empty string, etc.

Ticket: ENT-10944

### DIFF
--- a/cfbs/build.py
+++ b/cfbs/build.py
@@ -18,6 +18,85 @@ from cfbs.utils import (
 from cfbs.pretty import pretty, pretty_file
 
 
+def validate_config_for_build_field(config):
+    """Validate that neccessary fields are in the config for the build/download commands to work"""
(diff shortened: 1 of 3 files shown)


oleorhagen (approved):

larsewi (approved):
Nice :rocket: Should also be added to the `cfbs validate` command?

olehermanse:
@larsewi not in the current state. This validation is for the things needed for `cfbs build` to work. I guess we want `cfbs validate` to work for all `cfbs.json` files, so we'd have to make it less strict / more smart.

PR #174, opened 2024-01-09
Misc build key fixes on file without a build key

oleorhagen:


oleorhagen:
Based on top of https://github.com/cfengine/cfbs/pull/173 Didn't bother moving them from the same branch, so ignore the first three commits

JSON.md
```diff
@@ -29,11 +29,24 @@ As an example, here is a local policy file module with 3 common build steps:
...
+      "commit": "ca637d4e6148432a90b7db598a4137956c0e0282",
+      "added_by": "cfbs add",
+      "steps": [
+        "copy ./policy.cf services/cfbs/policy.cf",
+        "policy_files ./services/cfbs/policy.cf",
+        "bundles my_bundle"
+      ]
+    }
+  ],
+  "git": false
```
olehermanse, line 49: This example is not a full `cfbs.json` file, just one module. We can add a clarifying sentence about this, but this change doesn't make much sense. (`tags` and `added_by` should be at the module level, not the project / `cfbs.json` top level, the name of a custom policy file should be something like it was `./policy.cf`, and it should not have version, repo, commit SHA, etc.).

olehermanse (approved):

JSON.md
olehermanse: https://github.com/cfengine/cfbs/pull/176

PR #175, opened 2024-01-09
Added back validation of modules in "build" if present

olehermanse:
- Renamed parameter require_build_actions
- Added back validation of modules in "build" if present


larsewi (approved):

oleorhagen (approved):
Makes sense to me

PR #181, opened 2024-01-11
Always run pretty with the default sort conf when saving cfbs.json

oleorhagen:


olehermanse (approved):
Looks good to me, thanks.

olehermanse:
We could add a basic test for this. I tested locally, manually, and it seems to work correctly.

olehermanse:
@oleorhagen FYI, I added some more fixes / followups here: https://github.com/cfengine/cfbs/pull/182

PR #182, opened 2024-01-11
Fixed pretty formatting of first cfbs init commit

olehermanse:
- De-duplicated cfbs sorting rules
- Renamed sorting rule variables
- Fixed pretty formatting of first cfbs init commit
- Updated imports for top level keys and module keys

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -141,12 +141,14 @@ def init_command(index=None, masterfiles=None, non_interactive=False) -> int:
         default="Example description",
     )
 
-    config = {
-        "name": name,
-        "type": "policy-set",  # TODO: Prompt whether user wants to make a module
-        "description": description,
-        "build": [],
-    }
+    config = OrderedDict(
+        {
+            "name": name,
+            "type": "policy-set",  # TODO: Prompt whether user wants to make a module
+            "description": description,
+            "build": [],
+        }
+    )
     if index:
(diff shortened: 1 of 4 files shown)


craigcomstock (approved):

cfbs/commands.py
```diff
@@ -95,34 +100,7 @@ def pretty_command(filenames: list, check: bool, keep_order: bool) -> int:
     if not filenames:
         user_error("Filenames missing for cfbs pretty command")
 
-    cfbs_sorting_rules = None
```
oleorhagen, line 98: thx. forgot this one

cfbs/commands.py
```diff
@@ -207,7 +209,10 @@ def init_command(index=None, masterfiles=None, non_interactive=False) -> int:
 
     config["git"] = do_git
 
-    write_json(cfbs_filename(), config)
+    data = pretty(config, CFBS_DEFAULT_SORTING_RULES) + "\n"
```
oleorhagen, line 213: Q: Why not write it through the `cfbs_json` object like we do most other things?

olehermanse: Feel free to take a stab at it, shouldn't be **very** hard.

cfbs/commands.py
olehermanse: np

PR #188, opened 2024-01-24
Allowed numbers in namespaces in module input

olehermanse:
We wanted to make a `rhel_7_stig` namespace, which seems reasonable
so let's allow it.

I've seen other langauges not allowing identifiers to start with a number
to make things less ambiguous, so let's do that.

For example, `rhel_7_stig` is considered an acceptable namespace, while
`123`, `100_000`, and `1M` are not, since they look like values / numbers.


nickanderson (approved):
makes sense to me.

PR #194, opened 2024-05-28
Better prompts and defaults when adding by URL and adding masterfiles via cfbs init

olehermanse:
- **cfbs.json: added_by now documents when a masterfiles module was added by "cfbs init"**
- **cfbs init: Specifically add the masterfiles module from the masterfiles repo URL**
- **cfbs add: Now adds all modules from URL if user presses enter or passes --non-interactive**
- **cfbs add: Now skips the prompt if a repo added by URL only has 1 module**
- **cfbs add: Added information about dependencies when adding modules by URL**
- **cfbs add: Fixed issue printing empty list of modules before erroring when adding by URL**
- **cfbs add: Now prompts to add individual modules after user says no to add all from URL**


nickanderson:
SUBSCRIBED.

cfbs/commands.py
```diff
@@ -240,24 +240,26 @@ def init_command(index=None, masterfiles=None, non_interactive=False) -> int:
...
-            to_add = prompt_user(
-                non_interactive,
-                "Specify policy set to use instead (empty to skip)",
-                default="",
-            )
+            to_add = [
+                prompt_user(
+                    non_interactive,
+                    "Specify policy set to use instead (empty to skip)",
+                    default="",
```
nickanderson, line 257: Perhaps an example response. I guess, but am not sure that I should provide a module repo URL and perhaps a module name, maybe a commit sha in the URL?

nickanderson (approved):
Good changes IMHO. Suggestion to provide example input if answer no to default policy set question.

larsewi (approved):
Nice 🚀

PR #198, opened 2024-07-23
ENT-11437: `cfbs init` no longer displays a traceback when `git` is not found

jakub-nt:
Changelog: Title

Ticket: ENT-11437

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -53,6 +53,7 @@ from cfbs.internal_file_management import (
 )
 from cfbs.index import _VERSION_INDEX, Index
 from cfbs.git import (
+    check_git_exists,
     is_git_repo,
     git_commit,
     git_get_config,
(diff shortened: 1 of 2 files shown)


cfbs/commands.py
```diff
@@ -178,6 +179,12 @@ def init_command(index=None, masterfiles=None, non_interactive=False) -> int:
         do_git = True if do_git == "yes" else False
 
     if do_git is True:
+        try:
+            check_git_exists()
+        except FileNotFoundError as e:
+            print("Command 'git' was not found")
+            return 1
```
olehermanse, line 186: Looks like this will never happen?

From what I can see in the function, `check_git_exists()` raises a `CFBSGitError`, not a `FileNotFoundError`.

cfbs/git.py
```diff
@@ -18,6 +18,13 @@ class CFBSGitError(Exception):
     pass
 
 
+def check_git_exists():
+    try:
+        check_call(["git", "--version"], stdout=DEVNULL)
+    except CalledProcessError as cpe:
+        raise CFBSGitError("'git' was not found") from cpe
+
```
olehermanse, line 26: Currently, it looks like this function will only raise a `CFBSGitError`. If it indeed can raise `FileNotFoundError`, I would expect at least a comment mentioning this.

olehermanse (changes requested):
See comments

olehermanse (approved):

PR #199, opened 2024-07-25
ENT-11207: Building with invalid build steps now fails before executing any build steps, with an error message instead of a traceback

jakub-nt:


cf-bottom:

olehermanse (approved):
LGTM, thanks

larsewi:
Nice 🚀

PR #200, opened 2024-07-30
ENT-10229: Improved the warning text when an added local module contains a policy file with an autorun tag

jakub-nt:


cf-bottom:

larsewi (approved):
Great 🚀

cfbs/cfbs_config.py
```diff
@@ -286,13 +286,18 @@ def _handle_local_module(self, module):
             pattern = "%s/**/*.cf" % name
             policy_files = glob.glob(pattern, recursive=True)
 
+        modules_available = [m.get("name", "") for m in self.get("build", [])]
+        is_autorun_enabled = any(
+            m in modules_available for m in ["autorun", "./def.json"]
+        )  # only a heuristic
+
         for file in policy_files:
             if _has_autorun_tag(file):
```
olehermanse, line 295: Could `and` here instead of nested `if`s.

jakub-nt: This was on purpose, as the comment at https://github.com/cfengine/cfbs/pull/200/files#diff-16156b18d8d412a896d384b37f3947253ba5f775b6357a7205e0ef533d776d53R301 applies to the first condition, but not the second.

PR #201, opened 2024-07-31
ENT-10830: Build steps no longer write duplicate entries to `def.json` where appropriate

jakub-nt:


olehermanse (changes requested):
Looks good, as noted in Slack, please add a sentence or two about this in `JSON.md`.

larsewi (approved):
Very nice :rocket:

JSON.md
```diff
@@ -263,6 +263,8 @@ Unless otherwise noted, all steps are run inside the module's folder (`out/steps
   * Source is relative to module directory and target is relative to `out/masterfiles`.
     * In most cases, the build step should be: `input ./input.json def.json`
 
+If `def.json` is modified during a build step, the values of some of its keys are deduplicated. These keys are: `"inputs"`, `"augments"`, the `"tags"` key of each variable in the `"variables"` key, the `"class_expressions"` and `"tags"` keys of each class in the `"classes"` key if the class is a dictionary, and the class's own key if the class is a list.
```
olehermanse, line 266:
```suggestion
When `def.json` is modified during a `json`, `input`, `directory`, `bundles`, or `policy_files` build step, the values of some lists of strings are deduplicated, when this does not make any difference in behavior.
These cases are:

1. Policy files and augments files in the `"inputs"` and `"augments"` top level keys.
2. `"tags"` inside variables in `"variables"` and classes in `"classes"`.
3. Class expressions for each class in `"classes"`.
   These are in the subkey `"class_expressions"` when the class is defined using an object, and if the class is defined using just a list, that list is the list of class expressions implicitly.
```

olehermanse (changes requested):
Looks good in general, just one suggestion to improve the documentation.

nickanderson (changes requested):
Looks pretty good to me. But the first commit description could be improved, or perhaps better squash this all into a single commit.

PR #205, opened 2024-08-20
CFE-3891: Initial policy set modules are now correctly treated like manually added modules, rather than dependencies

jakub-nt:


cf-bottom:
Thanks for submitting a PR! Maybe @larsewi can review this?

cfbs/module.py
```diff
@@ -3,6 +3,10 @@
 from cfbs.pretty import pretty
 
 
+def is_module_added_manually(added_by: str):
+    return (added_by == "cfbs add") or (added_by == "cfbs init")
```
larsewi, line 7: You could consider
```suggestion
    return added_by in ("cfbs add", "cfbs init")
```
but there is nothing wrong with what you did ;)

cfbs/commands.py
```diff
@@ -521,7 +521,7 @@ def _clean_unused_modules(config=None):
         return 0
 
     def _someone_needs_me(this) -> bool:
-        if "added_by" not in this or this["added_by"] == "cfbs add":
+        if "added_by" not in this or is_module_added_manually(this["added_by"]):
```
larsewi, line 524: Operator precedence can be a bit confusing sometimes. Hence, adding parentheses around could help the readability
```suggestion
        if ("added_by" not in this) or is_module_added_manually(this["added_by"]):
```
But feel free to dismiss, if you disagree.

larsewi (approved):
Looks great :rocket:

olehermanse (approved):

PR #208, opened 2024-10-17
ENT-12098: Added an initial implementation of an internal cfbs command to generate MPF release information

jakub-nt:


cf-bottom:
Thank you for submitting a PR! Maybe @craigcomstock can review this?

masterfiles/analyze.py
```diff
@@ -0,0 +1,54 @@
+# TODO merge this with ENT-12099 branch cfbs analyze.py
+import os
```
olehermanse, line 2: Please put these files inside the `cfbs` package is installed as one self contained package. (`./cfbs/masterfiles/analyze.py`, `./cfbs/masterfiles/download_all_versions.py`, etc.)

cfbs/utils.py
```diff
@@ -278,6 +286,18 @@ def cfbs_dir(append=None) -> str:
...
+def string_sha256(input):
+    return hashlib.sha256(input.encode("utf-8")).hexdigest()
+
+
+def file_sha256(file):
+    h = hashlib.sha256()
+
+    h.update(open(file, "rb").read())
+
+    return h.hexdigest()
```
olehermanse, line 298: Nice utility functions. Not very critical in this case, but you could add some unit tests to `./tests/test_utils.py`

cfbs/commands.py
```diff
@@ -65,6 +65,7 @@
 from cfbs.git_magic import Result, commit_after_command, git_commit_maybe_prompt
 from cfbs.prompts import YES_NO_CHOICES, prompt_user
 from cfbs.module import Module, is_module_added_manually
+from masterfiles.generate_release_information import generate_release_information
```
olehermanse, line 68: When masterfiles is inside cfbs, it'll look like this:

```suggestion
from cfbs.masterfiles.generate_release_information import generate_release_information
```

masterfiles/check_download_matches_git.py
```diff
@@ -0,0 +1,20 @@
+# check that the downloadable files match the git files, mitigating a build system supply-chain attack
```
olehermanse, line 1: This looks like a good case for a docstring. Also it's not just about supply chain attack, I'd go with a bit more generic description:

```suggestion
"""Check that the downloadable files match the git files

This can be used to monitor / detect if something has been changed, accidentally or maliciously.
"""
```

olehermanse (commented):
Looks good so far, some preliminary comments.

cfbs/utils.py
jakub-nt: Done in https://github.com/cfengine/cfbs/pull/208/commits/2768a80ff7341b1d1e7330180f1f5e66ac771605

cfbs/commands.py
jakub-nt: Done in https://github.com/cfengine/cfbs/pull/208/commits/3f0f1ed42b9971a87d0bf816fe3e9ed416fd7371

masterfiles/analyze.py
jakub-nt: Done in https://github.com/cfengine/cfbs/pull/208/commits/3f0f1ed42b9971a87d0bf816fe3e9ed416fd7371

cfbs/masterfiles/analyze.py
```diff
@@ -0,0 +1,73 @@
+# TODO merge this with ENT-12099 branch cfbs analyze.py
+import os
+
+from cfbs.utils import dict_sorted_by_key, file_sha256
+
+IGNORED_PATH_COMPONENTS = [".git/", ".gitignore", ".gitattributes"]
+# ignore a path iff it contains a component (single file or directory) from this list
```
craigcomstock, line 7:
```suggestion
# ignore a path if it contains a component (single file or directory) from this list
```

cfbs/masterfiles/analyze.py
```diff
@@ -0,0 +1,73 @@
...
+# an element of this list should be just one component
+# folders should end with '/', files should not
+# TODO
```
craigcomstock, line 10: What is the TODO? The stuff above? I usually put the `TODO` before the description of the todo.

cfbs/masterfiles/generate_vcf_git_checkout.py
```diff
@@ -0,0 +1,92 @@
...
+    if os.path.isdir(MPF_PATH):
+        shutil.rmtree(MPF_PATH)
+
+    subprocess.run(
+        ["git", "clone", MPF_URL],
+        cwd=DIR_PATH,
+        check=True,
+    )
+
+    if not git_exists():
```
craigcomstock, line 32: Why is the check for git exists after a use of git? Seems like it should be up around line 25.

cfbs/masterfiles/generate_vcf_git_checkout.py
```diff
@@ -0,0 +1,92 @@
...
+        # a clean is necessary to remove all the undesired files
+        subprocess.run(
+            ["git", "clean", "-dffx"],
+            cwd=MPF_PATH,
+            check=True,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
+        )
+
+        # build masterfiles from git as they are in the tarball packages
```
craigcomstock, line 71: Seems you should check that autoconf and automake are installed first so that autogen.sh can work.

craigcomstock: Since you are using `check=True` with `run()` but not catching any exceptions I worry about hard to read stack traces in case of problems. Probably fine in the exceptional cases but something like autoconf missing should be checked I think as that seems likely to happen.

craigcomstock (changes requested):
Cool. Noticed a few things with a cursory look. I didn't check the logic too carefully. I can make another pass if need be.

cfbs/masterfiles/analyze.py
jakub-nt: This one is on purpose, "iff" means "if and only if".

cfbs/masterfiles/analyze.py
jakub-nt: Yes, sure, I'll change this.

cfbs/masterfiles/generate_vcf_git_checkout.py
jakub-nt: Oops, that's an oversight, in an earlier version checking in this place was sufficient. Will fix.

cfbs/masterfiles/analyze.py
jakub-nt: Reworded in https://github.com/cfengine/cfbs/pull/208/commits/b516b19c038d149e10df342e60b71cb777cc34b5

cfbs/masterfiles/analyze.py
jakub-nt: Rewritten in https://github.com/cfengine/cfbs/pull/208/commits/b516b19c038d149e10df342e60b71cb777cc34b5

cfbs/masterfiles/generate_vcf_git_checkout.py
jakub-nt: Fixed in https://github.com/cfengine/cfbs/pull/208/commits/b516b19c038d149e10df342e60b71cb777cc34b5

cfbs/masterfiles/generate_vcf_git_checkout.py
jakub-nt: Added checking for required commands in https://github.com/cfengine/cfbs/pull/208/commits/b516b19c038d149e10df342e60b71cb777cc34b5.

As for the stack traces, I've checked a few cases and I think it's readable enough. For the git command calls, the failing command and the status code should be enough information, and for the autogen and make calls, currently the entire stdout and stderr are printed.

craigcomstock (approved):
Nice.

cfbs/masterfiles/generate_vcf_git_checkout.py
```diff
@@ -0,0 +1,101 @@
...
+
+def check_required_command(command):
+    if not shutil.which(command):
+        print("`%s` was not found" % command)
+        sys.exit(1)
+
+
+def check_required_commands(commands):
+    for c in commands:
+        check_required_command(c)
```
larsewi, line 28: These two can probably be the same function. If the passed argument is an interable, then it can recursively call itself for each element in the iterable.

cfbs/masterfiles/generate_vcf_git_checkout.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def generate_vcf_git_checkout(interesting_tags=None):
+    required_commands = ["git", "make", "automake", "autoconf"]
+    check_required_commands(required_commands)
+
+    # clone the MPF repo every time the script is run, in case there are updates
```
larsewi, line 35: Can you not rebase instead?

cfbs/masterfiles/generate_vcf_git_checkout.py
```diff
@@ -0,0 +1,101 @@
...
+    if os.path.isdir(MPF_PATH):
+        shutil.rmtree(MPF_PATH)
+
+    subprocess.run(
```
larsewi, line 39: Maybe we should handle exceptions if the command fails? I'm not sure.

cfbs/masterfiles/generate_vcf_git_checkout.py
```diff
@@ -0,0 +1,101 @@
...
+        interesting_tags = []
+
+        for tag in tags:
+            if "-" not in tag:
+                interesting_tags.append(tag)
+
+    versions_dict, checksums_dict, files_dict = initialize_vcf()
+
+    for tag in interesting_tags:
+        print(tag)
```
larsewi, line 61: Leftover debug print?

larsewi (changes requested):
Great work :rocket: I see you still have some `# TODO`'s left to address. I have some smaller nitpicks for you. But in general:
- It's nice if you try to avoid too many nested `if`'s. They then to make it really hard to understand the code. But this is not always the case.
- When you look for the masterfiles tarball URL, maybe it's possible to search for it, instead of hard coding where it is in the JSON? Maybe there is a common denominator you can look for? I'm not sure.

I tried running your program, but it hung. Or maybe it just takes a very long time? I did not try to figure out why.

Looking forward to re-review!

cfbs/masterfiles/download_all_versions.py
```diff
@@ -0,0 +1,86 @@
+from pathlib import Path
+import shutil
+import urllib.request
+
+from cfbs.utils import get_json
+
+DOWNLOAD = True
```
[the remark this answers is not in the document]
jakub-nt: I'll just remove it.

cfbs/masterfiles/check_tarball_checksums.py
```diff
@@ -0,0 +1,34 @@
+from cfbs.utils import file_sha256, immediate_files
+
+
+def check_tarball_checksums(dir_path, downloaded_versions, reported_checksums):
+    does_match = True
+
+    for version in downloaded_versions:
+        print(version)
+
+        version_path = dir_path / version
```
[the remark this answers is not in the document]
jakub-nt: Yes, this is an operator and it's `pathlib`'s equivalent of `os.path.join`. I'll just move everything to `os.path`.

cfbs/masterfiles/generate_vcf_git_checkout.py
```diff
@@ -0,0 +1,106 @@
...
+        subprocess.run(["make"], cwd=MPF_PATH, check=True)
+
+        # compute VCF data for all the files
+        versions_dict, checksums_dict, files_dict = versions_checksums_files(
+            MPF_PATH, tag, versions_dict, checksums_dict, files_dict
+        )
+
+        # clean the files to prevent spillage to other versions
+        subprocess.run(
+            ["git", "clean", "-dffx"],
```
larsewi, line 93: I don't think the second `f` is necessary here, since we don't do `git clone --recursive <REPO>`. However, it should not hurt either. So feel free to dismiss.
```suggestion
            ["git", "clean", "-dfx"],
```

larsewi (approved):
Thank you, this looks good to me :rocket:

cfbs/utils.py
```diff
@@ -278,6 +308,18 @@ def cfbs_dir(append=None) -> str:
...
 
 
+def string_sha256(input):
+    return hashlib.sha256(input.encode("utf-8")).hexdigest()
+
+
+def file_sha256(file):
+    h = hashlib.sha256()
+
+    h.update(open(file, "rb").read())
```
larsewi, line 318: Remember to `close()` the file afterwards. Or you can use the `with` keyword to do it automatically

larsewi (commented):
Great improvements :rocket: I have some minor comments and some open questions that I hope @olehermanse or @craigcomstock can answer

cfbs/masterfiles/check_download_matches_git.py
```diff
@@ -0,0 +1,46 @@
...
+            only_dl, only_git, value_diff = dict_diff(
+                download_version_dict, git_version_dict
+            )
+
+            print("Files only in the downloaded version:", only_dl, file=f)
+            print("Files only in the git version:", only_git, file=f)
+            print("Files with different contents:", value_diff, file=f)
+
+            if len(only_dl) > 0 or len(value_diff) > 0:
+                user_error(
```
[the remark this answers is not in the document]
craigcomstock: I think user_error() which calls sys.exit() makes sense here. Maybe the message should suggest what to do!

What action would we take if the files don't match git? Fix git? Fix the downloadable files? Fix something in this codebase that understands some quirk of a particular release?

cfbs/masterfiles/generate_vcf_git_checkout.py
```diff
@@ -0,0 +1,113 @@
...
+            check=True,
+        )
+
+    versions_dict, checksums_dict, files_dict = initialize_vcf()
+
+    for tag in checkout_tags:
+        print("Checkouting tag", tag)
+
+        # checkouting some tags equal to the downloaded version doesn't result in the same files
+        # the downloadable files are reproducible by checkouting specific tags
```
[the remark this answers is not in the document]
craigcomstock: I might suggest that we FIX these tags that are exceptions here. It would seem of not much consequence since all of these versions are out of support.

cfbs/masterfiles/generate_vcf_git_checkout.py
```diff
@@ -0,0 +1,113 @@
...
+            release_number = "1"
+        subprocess.run(
+            ["./autogen.sh"],
+            cwd=MPF_PATH,
+            check=True,
+            env=dict(
+                os.environ.copy(), EXPLICIT_VERSION=tag, EXPLICIT_RELEASE=release_number
+            ),
+        )
+        # older masterfiles version READMEs instruct to use `make install` and newer `make` - always use `make` instead
```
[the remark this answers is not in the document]
craigcomstock: I do think that `make dist` makes a lot of sense to me.

cfbs/utils.py
olehermanse: +1 for using context managers (`with`)

larsewi (approved):
LGTM :rocket: @jakub-nt maybe you can squash the commits? @olehermanse do you want to have one last look before we hit the merge button?

cfbs/masterfiles/generate_vcf_git_checkout.py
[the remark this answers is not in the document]
olehermanse: It could be something to do in addition to, not instead of. We explicitly want to compare what's in git here.

olehermanse (approved):
👍 Looks good to me, merging, and we can always submit follow-up fixes in new PRs.

cfbs/masterfiles/check_download_matches_git.py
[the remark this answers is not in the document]
jakub-nt: https://github.com/cfengine/cfbs/pull/212 makes the change to no longer exit on the first differing version.

cfbs/masterfiles/download_all_versions.py
```diff
@@ -0,0 +1,117 @@
...
+        mkdir(version_path)
+
+        # download a version, and verify the reported checksum matches
+        filename = url.split("/")[-1]
+        tarball_path = os.path.join(version_path, filename)
+        checksum = reported_checksums[version]
+        try:
+            fetch_url(url, tarball_path, checksum)
+        except FetchError as e:
+            user_error("For version " + version + ": " + str(e))
```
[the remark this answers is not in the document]
jakub-nt: https://github.com/cfengine/cfbs/pull/208#discussion_r1862058268

PR #212, opened 2024-11-27
ENT-12486: Output download and git differences to a JSON file, and do not exit on the first difference

jakub-nt:


cf-bottom:
Thanks for submitting a pull request! Maybe @craigcomstock can review this?

cfbs/masterfiles/check_download_matches_git.py
```diff
@@ -29,18 +34,64 @@ def check_download_matches_git(versions):
...
+        for filepath in only_git:
+            version_diffs_dict["files_only_in_git"].append(filepath)
+        for filepath, _, _ in value_diff:
+            version_diffs_dict["files_with_different_content"].append(filepath)
+
+        diffs_dict["differences"][version] = version_diffs_dict
+
+        if len(only_dl) > 0 or len(value_diff) > 0:
+            nonmatching_versions.append(version)
+            extraneous_count += len(only_dl)
```
larsewi, line 57: Does files existing only in git also belong in this count or maybe some other count that also can be displayed?

cfbs/masterfiles/check_download_matches_git.py
```diff
@@ -29,18 +34,64 @@ def check_download_matches_git(versions):
...
+        for filepath in only_dl:
+            version_diffs_dict["files_only_in_downloads"].append(filepath)
+        for filepath in only_git:
+            version_diffs_dict["files_only_in_git"].append(filepath)
+        for filepath, _, _ in value_diff:
+            version_diffs_dict["files_with_different_content"].append(filepath)
+
+        diffs_dict["differences"][version] = version_diffs_dict
+
+        if len(only_dl) > 0 or len(value_diff) > 0:
```
larsewi, line 55: Here you can consider:
```suggestion
        if only_dl or value_diff:
```
Since an empty list evaluates to `False` and non-empty evaluates to `True`. E.g.:
```Python
>>> bool([])
False
>>> bool(["foo"])
True
```
But you could argue that the former is more readable.

larsewi (approved):
LGTM :rocket:

cfbs/masterfiles/check_download_matches_git.py
jakub-nt: Yes, this was on purpose for readability.

cfbs/masterfiles/check_download_matches_git.py
jakub-nt: git-only count is not productive currently, it would need work outside the scope of this PR.

PR #215, opened 2024-12-02
Add argument to specify minimal version in `cfbs generate-release-information`

jakub-nt:


cf-bottom:
Thanks for submitting a pull request! Maybe @larsewi can review this?

cfbs/masterfiles/analyze.py
```diff
@@ -120,3 +120,9 @@ def version_as_comparable_list_negated(version):
     vcl[1] = [-x for x in vcl[1]]
 
     return vcl
+
+
+def version_is_at_least(version, min_version):
```
larsewi, line 125: It would be nice to have a couple of versions compare functions in `utils.py` for the different comparison operators `>` , `<` , `>=` , `<=` , `==` , and `!=` along with unit tests. This is probably out of the scope of this PR. But feel free to do that in a follow-up PR, if you are up for the challenge :wink: I believe we compare versions in multiple places throughout this project. `cfbs add` is one example.

cfbs/masterfiles/generate_release_information.py
```diff
@@ -7,16 +8,23 @@
...
         print("Downloading masterfiles...")
 
-        downloaded_versions = download_all_versions(DOWNLOAD_PATH)
+        downloaded_versions = download_all_versions(DOWNLOAD_PATH, min_version)
 
         print("Download finished. Every reported checksum matches.")
     else:
         downloaded_versions = immediate_subdirectories(DOWNLOAD_PATH)
 
+        downloaded_versions = list(
```
larsewi, line 21: What happens if you omit the `list()` function? I suspect it might work just fine.

larsewi (commented):
Some smaller comments, otherwise LGTM :rocket:

cfbs/masterfiles/generate_release_information.py
jakub-nt: I think it's better to keep it so that things like `print` work as expected without modifications.

larsewi (approved):

PR #218, opened 2025-01-22
[ENT-12099] Initial implementation of an experimental feature of policy set analysis

jakub-nt:


cf-bottom:
Thanks for submitting a PR! Maybe @craigcomstock can review this?

cfbs/analyze.py
```diff
@@ -0,0 +1,521 @@
...
+        "downloads/github.com/cfengine/release-information/archive/refs/heads/"
+        + RI_SHA1_CHECKSUM
+        + "/release-information-main/masterfiles/"
+    )
+    mpf_vcf_path = os.path.join(cfbs_dir(), mpf_vcf_subpath)
+    if not os.path.exists(mpf_vcf_path):
+        fetch_archive(RI_ARCHIVE_URL, RI_SHA1_CHECKSUM, with_index=False)
+    # TODO the release information checksum needs to be updated on each new release
+    # currently, if the checksum is not manually updated, the old, already downloaded files will continue to be used
+    # and if the old files have not already been downloaded, the download from GitHub will fail
```
craigcomstock, line 98: Right. So when we release we need to update this.
We need to add instructions to this template ticket for releases: https://northerntech.atlassian.net/browse/ENT-12575

cfbs/analyze.py
```diff
@@ -0,0 +1,521 @@
...
+    mpf_files_dict = get_or_read_json(mpf_files_json_path)
+    mpf_files_dict = mpf_files_dict["files"]
+
+    return mpf_versions_dict, mpf_checksums_dict, mpf_files_dict
+
+
+def filepath_sorted(filepaths):
+    """Returns a list of file paths sorted hierarchically.
+
+    Currently not implemented, sorting regularly instead, TODO."""
```
craigcomstock, line 119: Make a ticket and reference it here? How important is sorting it hierarchically?

cfbs/analyze.py
```diff
@@ -0,0 +1,521 @@
...
+                reference_version_checksums[mpf_checksum] = []
+            reference_version_checksums[mpf_checksum].append(mpf_filepath)
+
+    analyzed_files = AnalyzedFiles(reference_version)
+
+    # categorize all files, based on their relation with the reference version and known MPF files:
+    # 1. files present:
+    for checksum, files_of_checksum in checksums_dict.items():
+        if checksum in mpf_checksums_dict:
+            # 1A. checksum known:
```
craigcomstock, line 471: I am a little confused by the duplicate 1A, 1A1, 1A1A etc here. I know the comments aren't functional but the duplication seems like it would make it harder to refer/search to a particular section.

craigcomstock (commented):
a few questions/comments, all looks good-to-go except maybe moving the release-information checksum into a separate file, that might be a nice change for maintenance purposes.

cfbs/analyze.py
olehermanse: @craigcomstock @jakub-nt no, we should not do it like this. As it's implemented here, `cfbs analyze` essentially stops working in all versions of cfbs when we release something new in release-information, and we have to release a new version of cfbs to fix it.

Keep in mind that HTTPS already has signatures and thus integrity checking, and so should be protected against corrupt downloads / MITM attacks. So what we are doing here is an additional check, to protect against for example a compromised user having access to edit the data in GitHub.

The correct way to do this is via some form of "releases". Either GH releases, with a download zip and a checksums.txt - assuming that they are a bit better protected, i.e. fewer people have access to edit releases. Or to avoid the single point of failure in GitHub, we could put the checksums.txt in another place, like our cfengine.com website, via the releases.json.

cfbs/analyze.py
craigcomstock: Ideally, I think you could refactor each branch label into a function that is shared and that might improve the readability of this section which has quite a deep nest of indents which is often a sign of need for refactoring I would say.

craigcomstock: I think this is fine as-is. The code works and is 80% readable. A refactor of the same labels as functions would probably improve readability but at this point I don't think it is critical at all.

cfbs/analyze.py
```diff
@@ -0,0 +1,597 @@
...
+        self.version_counter = VersionsCounter()
+        self.highest_version_counter = VersionsCounter()
+        # acronyms: vc = version_counter, hvc = highest_version_counter
+        self.different_filepath_vc = VersionsCounter()
+        self.different_filepath_hvc = VersionsCounter()
+
+    def display(self):
+        if not self.version_counter.is_empty():
+            print(
+                "Same filepath versions distribution:",
```
larsewi, line 267: From looking at just the output from the command, it is difficult to understand what this is

cfbs/analyze.py
```diff
@@ -0,0 +1,597 @@
...
+                self.version_counter.sorted_list(),
+            )
+            print(
+                "Same filepath highest versions distribution:",
```
larsewi, line 271: Same here

larsewi (approved):
LGTM :rocket:

cfbs/args.py
```diff
@@ -115,6 +115,31 @@ def get_arg_parser():
...
+        "--reference-version",
+        help="Specify version to compare against for 'cfbs analyze'",
+    )
+    parser.add_argument(
+        "--masterfiles-dir",
+        help="If the path given to 'cfbs analyze' contains a masterfiles subdirectory, specify the subdirectory's name",
+    )
+    parser.add_argument(
+        "--ignored-path-components",
+        help="Specify path components which should be ignored during 'cfbs analyze'",
```
craigcomstock, line 135: The help should probably mention that more than one component is specified with spaces like `--ignored-path-components one two three`.

cfbs/cfbs.1
```diff
@@ -85,6 +85,22 @@ Check whether masterfiles from cfengine.com and github.com match in 'cfbs genera
...
+\fB\-\-reference\-version\fR \fI\,REFERENCE_VERSION\/\fR
+Specify version to compare against for 'cfbs analyze'
+
+.TP
+\fB\-\-masterfiles\-dir\fR \fI\,MASTERFILES_DIR\/\fR
+If the path given to 'cfbs analyze' contains a masterfiles subdirectory, specify the subdirectory's name
+
+.TP
+\fB\-\-ignored\-path\-components\fR \fI\,[IGNORED_PATH_COMPONENTS ...]\/\fR
+Specify path components which should be ignored during 'cfbs analyze'
```
craigcomstock, line 102: also mention space-delimited list for more than one here?

cfbs/commands.py
```diff
@@ -1057,6 +1057,72 @@ def info_command(modules):
...
+def analyze_command(
+    policyset_paths,
+    json_filename=None,
+    reference_version=None,
+    masterfiles_dir=None,
+    user_ignored_path_components=None,
+    offline=False,
+):
+    if len(policyset_paths) == 0:
+        # no policyset path is a shorthand for using the current directory as the policyset path
```
craigcomstock, line 1071: Maybe instead of a comment here add an info or verbose message?

cfbs/analyze.py
```diff
@@ -0,0 +1,597 @@
...
+            checksums_dict["checksums"][file_checksum].add(tarball_relpath)
+
+            if not tarball_relpath in files_dict["files"]:
+                files_dict["files"][tarball_relpath] = set()
+            files_dict["files"][tarball_relpath].add(file_checksum)
+
+    return checksums_dict, files_dict
+
+
+def mpf_vcf_dicts(offline=False):
```
craigcomstock, line 94: here I wonder if vcf at least needs to be expanded. MPF is probably OK.

cfbs/analyze.py
```diff
@@ -0,0 +1,597 @@
...
+    REPO_OWNER = "cfengine"
+    REPO_NAME = "release-information"
+
+    REPO_OWNERNAME = REPO_OWNER + "/" + REPO_NAME
+    RI_SUBDIRS = "downloads/github.com/" + REPO_OWNERNAME + "/archive/refs/tags/"
+
+    if offline:
+        ERROR_MESSAGE = "MPF release information not found. Provide the release information, for example by running 'cfbs analyze' without '--offline'."
```
craigcomstock, line 102: In a message to the user I wonder if MPF should be Masterfiles or Masterfiles Policy Framework.

cfbs/analyze.py
craigcomstock: Is there an example of the output here or in the man page contents?

craigcomstock (commented):
Generally looks great.

nickanderson (commented):
@jakub-nt some thoughts from playing with this recently:

-   Would be cool to provide an option to show the diff. Would have to have the actual original file to show that &#x2026;

```
Files from the version but with modifications:
└── masterfiles/promises.cf
```

-   Convert to cfbs, interactively

-   Condense these?

```
Files missing from the version:
└── masterfiles/lib/3.7/paths.cf
Files moved or renamed:
└── ('masterfiles/lib/3.7/paths.cf-moved', [('masterfiles/lib/3.7/paths.cf', ['3.7.5'])])
```

Move some of the long dense messages with tuples to verbose logging.

cfbs/analyze.py
```diff
@@ -0,0 +1,597 @@
...
+        self.different_filepath_vc = VersionsCounter()
+        self.different_filepath_hvc = VersionsCounter()
+
+    def display(self):
+        if not self.version_counter.is_empty():
+            print(
+                "Same filepath versions distribution:",
+                self.version_counter.sorted_list(),
+            )
+            print(
```
olehermanse, line 270:
```suggestion
            log.debug(
```

cfbs/analyze.py
```diff
@@ -0,0 +1,597 @@
+from collections import OrderedDict
+import os
```
olehermanse, line 2:
```suggestion
import os
import logging as log
```

olehermanse:
Ye, the first 2 things Nick mention are separate tasks (diffs and convert). The last 2 things we probably want to fix. @jakub-nt let's do this:

1. Remove entries from the files missing section when they appear in the moved / renamed section.
2. Adjust the messages which are not so useful to an average user to `log.debug`, see review suggestions.

olehermanse (approved):

PR #222, opened 2025-04-09
ENT-12739: Added detection of the user providing wrong policy set path in cfbs analyze

jakub-nt:


cfbs/analyze.py
```diff
@@ -551,8 +558,51 @@ def analyze_policyset(
...
+                    possible_policyset_relpaths.append(relpath[:-10])
+                if relpath.endswith("/promises.cf"):
+                    possible_policyset_relpaths.append(relpath[:-12])
+
+        # for drive root, the path's parent is the path itself, so only check the parent path if this is not the case
+        if os.path.realpath(path) != os.path.realpath(os.path.join(path, "..")):
+            if os.path.exists(os.path.join(path, "..", "update.cf")) or os.path.exists(
+                os.path.join(path, "..", "promises.cf")
+            ):
+                possible_policyset_relpaths.append("..")
```
olehermanse, line 586: `drive root` is the root of the filesystem, right?

This part is a bit cryptic, and could be more readable, I think, something like this;

```suggestion
        # for file system root, the path's parent is the path itself, so only check the parent path if this is not the case
        current = os.path.realpath(path)
        parent = os.path.realpath(os.path.join(path, ".."))
        at_root = (path == "/") # or current == parent
        if not at_root:
            if os.path.exists(os.path.join(parent, "update.cf")) or os.path.exists(
                os.path.join(parent, "promises.cf")
            ):
                possible_policyset_relpaths.append("..")
```

That might even make the comment unnecessary (which is a good thing).

Also, importing the functions you need from `os.path` could help reduce noise;

```
from os.path import exists, realpath, join
```

olehermanse (approved):

PR #226, opened 2025-06-04
CFE-3847: Made possible to update modules added by url

victormlg:
Added a way to update modules added by url. I tried to do it with minimal extra code inside the update loop. I am not sure if we want to delete the older `commit/repo`


cfbs/commands.py
```diff
@@ -677,173 +788,117 @@ def update_command(to_update):
...
-        custom_index = module is not None and "index" in module
-        index = Index(module["index"]) if custom_index else config.index
+        custom_index = old_module is not None and "index" in old_module
+        index = Index(old_module["index"]) if custom_index else config.index
 
-        if not module:
+        if not old_module:
             index.translate_alias(update)
-            module = config.get_module_from_build(update.name)
+            old_module = config.get__module_from_build(update.name)
```
jakub-nt, line 802: Typo (double underscore), errors on run.

jakub-nt (dismissed):

cfbs/commands.py
```diff
@@ -663,6 +665,115 @@ def _update_variable(input_def, input_data):
...
+            if commit_differs:
+                old_module["input"] = new_module["input"]
+
+                input_path = os.path.join(".", old_module["name"], "input.json")
+                input_data = read_json(input_path)
+                if input_data == None:
+                    log.debug(
+                        "Skipping input update for module '%s': "
+                        + "No input found in '%s'" % (old_module["name"], input_path)
+                    )
```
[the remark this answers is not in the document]
olehermanse: Adding parentheses is also an option (around the plussed strings).

tests/shell/042_update_from_url/delete-files/input.json
```diff
@@ -0,0 +1,25 @@
...
+            "label": "Path",
+            "question": "Path to file"
+        },
+        {
+            "key": "why",
+            "type": "string",
+            "label": "Why",
+            "question": "Why should this file be deleted?",
+            "default": "Unknown"
+        }
```
larsewi, line 21: Here and in cfbs.json you can maybe remove this object causing the input specification to change. This way we can pretend that this string field was added in the "newer" version of the module

olehermanse (approved):

PR #228, opened 2025-06-06
CFE-4537, CFE-4538: more unit test coverage

jakub-nt:


cfbs/utils.py
```diff
@@ -280,7 +281,6 @@ def immediate_files(path):
 
 
 def path_append(dir, subdir):
-    dir = os.path.abspath(os.path.expanduser(dir))
```
olehermanse, line 283: Same as above, and below in `cfengine_dir()`. Avoid changing the implementation, add only tests first, and then introduce a separate PR, with individual commits explaining why the change is necessary and correct.

cfbs/utils.py
```diff
@@ -296,7 +296,10 @@ def are_paths_equal(path_a, path_b) -> bool:
 
 
 def cfengine_dir(subdir=None):
-    return path_append("~/.cfengine/", subdir)
+    CFENGINE_DIR = "~/.cfengine/"
+    cfengine_dir_abspath = os.path.abspath(os.path.expanduser(CFENGINE_DIR))
+
+    return path_append(cfengine_dir_abspath, subdir)
```
olehermanse, line 302: Same here.

jakub-nt:
Closing in favor of #233, #234 and #235.

PR #230, opened 2025-06-18
pretty_file: Added detection whether reformatting is necessary

olehermanse:


tests/test_pretty.py
```diff
@@ -205,6 +205,15 @@ def test_pretty_string():
     assert pretty_string(test) == expected
 
 
+def test_pretty_file():
```
larsewi, line 208: You could use the `tmp_path` fixture in `pytest` (see https://docs.pytest.org/en/6.2.x/tmpdir.html)

PR #232, opened 2025-06-20
CFE-3841: Do not display skipping module add messages if the module is not specified by the user

jakub-nt:
Depends on #228.


olehermanse (changes requested):
Please rebase now that the related PR is merged.

PR #234, opened 2025-06-23
CFE-4538: Add more unit test coverage

jakub-nt:

PR #235, opened 2025-06-23
Minor code quality improvements

jakub-nt:
Depends on #234.


JSON.md
```diff
@@ -222,7 +222,7 @@ These are copies of the module directories, where it's more "safe" to do things
 ## All available build steps
 
 The build steps below manipulate the temporary files in the steps directories and write results to the output policy set, in `out/masterfiles`.
-Unless otherwise noted, all steps are run inside the module's folder (`out/steps/...`) with sources / file paths relative to that folder, and targets / destinations mentioned below are relative to the output policy set (`out/masterfiles`, which in the end will be deployed as `/var/cfengine/masterfiles`).
+Unless otherwise noted, all steps are run inside the module's folder (`out/steps/...`) with sources / file paths relative to that folder, and targets / destinations mentioned below are relative to the output policy set (`out/masterfiles`, which in the end will be deployed as `/var/cfengine/masterfiles`). In `cfbs.json`'s `"steps"`, the build step name must be separated from the rest of the build step by a regular space.
```
olehermanse, line 225: Looks like the rest of the file is using our recommended 1 sentence per line style - should follow this here as well, it would have made the diff nicer as well.

```suggestion
Unless otherwise noted, all steps are run inside the module's folder (`out/steps/...`) with sources / file paths relative to that folder, and targets / destinations mentioned below are relative to the output policy set (`out/masterfiles`, which in the end will be deployed as `/var/cfengine/masterfiles`).
In `cfbs.json`'s `"steps"`, the build step name must be separated from the rest of the build step by a regular space.
```

cfbs/utils.py
```diff
@@ -295,7 +294,6 @@ def immediate_files(path):
 
 
 def path_append(dir, subdir):
-    dir = os.path.abspath(os.path.expanduser(dir))
```
olehermanse, line 286: Here, your commit message should include more information on why. It looks like you're changing the implementation just to make it nicer to test. It could be useful that this function always "normalizes" paths like this. I see that it's only used once in this codebase, so probably doesn't matter too much, but it's also in cf-remote (copy-pasted) where it's used a bit more.

olehermanse (changes requested):
Please rebase now that the related PRs are merged :)

olehermanse (approved):

PR #236, opened 2025-06-23
JSON.md: Reformatted with Prettier

olehermanse:
We've used prettier in other places for markdown files and it works nicely.


jakub-nt (approved):

nickanderson:
Very nice

PR #239, opened 2025-06-25
CFE-4040: Added a user error on add from a URL of an unsupported scheme, and fixed crash on attempting to add a non-URL archive file

jakub-nt:


cfbs/cfbs_config.py
```diff
@@ -386,16 +386,20 @@ def add_command(
 
         before = {m["name"] for m in self.get("build", [])}
 
-        if to_add[0].endswith(SUPPORTED_ARCHIVES) or to_add[0].startswith(
-            SUPPORTED_URI_SCHEMES
-        ):
+        if to_add[0].startswith(SUPPORTED_URI_SCHEMES):
```
larsewi, line 389: Should not the check to see if it ends with one of the `SUPPORTED_ARCHIVES` still be here?

larsewi (approved):

cfbs/cfbs_config.py
jakub-nt: If `to_add[0].endswith(SUPPORTED_ARCHIVES)` is true then either `to_add[0].startswith(SUPPORTED_URI_SCHEMES)` is true and nothing has changed, or it's false and `to_add[0]` isn't a supported URL and the subsequent code has no capability to add a local archive file. Actually neither `_add_using_url` nor `_add_modules` have such capability, and it seems adding local archive files has simply [never been supported](https://github.com/cfengine/cfbs/blob/8492e6a73603024b57a89f0ea276399336922ea1/cfbs/internal_file_management.py#L54) (we could consider supporting it as a local version of adding an archive file by URL in the future).

olehermanse (approved):

PR #242, opened 2025-07-02
Added new replace_version build step

olehermanse:


larsewi (approved):
Maybe the build step should be called only `replace` and made more generic? It could be useful for other things as well? Although I guess you could always use the `sh` build step in that case. 🤔

olehermanse:
> Maybe the build step should be called only replace and made more generic? It could be useful for other things as well? Although I guess you could always use the sh build step in that case. 🤔

Yes, I'm already working on a generic version, however I think it's better to have `replace_version` as well, mainly because it will have the magic behavior of finding the current version, rather than having to specify what should be put in.

PR #244, opened 2025-07-03
Added a utility function to strip any one of a list of strings

jakub-nt:


cfbs/utils.py
```diff
@@ -128,6 +128,24 @@ def strip_left(string, beginning):
     return string[len(beginning) :]
 
 
+def strip_right_any(string, suffixes: tuple):
+    if not string.endswith(suffixes):
+        return string
+
+    for suffix in suffixes:
+        if string.endswith(suffix):
+            return string[0 : -len(suffix)]
```
olehermanse, line 137:
```suggestion
def strip_right_any(string, suffixes: tuple) -> str:
    for suffix in suffixes:
        if string.endswith(suffix):
            return string[0 : -len(suffix)]
    return string
```

PR #247, opened 2025-07-04
ENT-13089: Added validation of module names

jakub-nt:


JSON.md
```diff
@@ -174,6 +174,9 @@ The modules inside `build`, `provides`, and `index` use these fields:
   For `provides` and `index` dictionaries, this name must be the key of each entry (not a field inside).
   For the `build` array, it must be inside each module object (with `name` as the key).
   Local modules (files and folders in same directory as `cfbs.json`), must start with `./`, and end with `/` if it's a directory.
+  Module names should not be longer than 64 characters.
+  Module names (not including adfixes `./`, `/`, `.cf`, `.json` for local modules) should only contain lowercase ASCII alphanumeric characters possibly separated by dashes, and should start with a letter.
```
olehermanse, line 178:
```suggestion
  Module names (not including affixes `./`, `/`, `.cf`, `.json` for local modules) should only contain lowercase ASCII alphanumeric characters possibly separated by dashes, and should start with a letter.
```

cfbs/utils.py
```diff
@@ -130,23 +130,21 @@ def strip_left(string, beginning):
     return string[len(beginning) :]
 
 
-def strip_right_any(string, suffixes: tuple):
-    if not string.endswith(suffixes):
-        return string
-
+def strip_right_any(string, suffixes):
```
olehermanse, line 133: Why remove the type? In this case, it looks like the most specific type is;

```suggestion
def strip_right_any(string: str, suffixes: Iterable[str]) -> str:
```

cfbs/utils.py
```diff
@@ -130,23 +130,21 @@ def strip_left(string, beginning):
...
         if string.endswith(suffix):
             return string[0 : -len(suffix)]
 
+    return string
 
-def strip_left_any(string, prefixes: tuple):
-    if not string.startswith(prefixes):
-        return string
 
+def strip_left_any(string, prefixes):
```
olehermanse, line 141: Same here.

cfbs/validate.py
```diff
@@ -166,6 +170,46 @@ def _validate_top_level_keys(config):
...
+            + " characters)",
+        )
+
+    # lowercase ASCII alphanumericals, starting with a letter, and possible singular dashes in the middle
+    r = "[a-z][a-z0-9]*(-[a-z0-9]+)*"
+    proper_name = name
+
+    if is_module_local(name):
+        if not name.startswith("./"):
+            raise CFBSValidationError(name, "Local module names should begin with `./`")
```
olehermanse, line 190: This is ensure by `is_module_local()`, so should be an assert / programmer error if that's not the case:

 ```suggestion
        assert name.startswith("./"), "is_module_local() should ensure this module starts with ./"
```

File: cfbs/validate.py [hunk 5]
```diff
@@ -166,6 +170,46 @@ def _validate_top_level_keys(config):
...
+        # allow underscores, only for local modules
+        proper_name = proper_name.replace("_", "-")
+
+    if not re.fullmatch(r, proper_name):
+        raise CFBSValidationError(
+            name,
+            "Module name contains illegal characters (only lowercase ASCII alphanumeric characters are legal)",
+        )
+
+    log.debug("Validated name of module %s" % name)
```
olehermanse, line 210:
```suggestion
    log.debug("Successfully validated name of module %s" % name)
```

olehermanse (approved):
Nice. Only small comments, feel free to fix in a follow-up PR.

PR #254, opened 2025-07-15
Fixed issues with repeated validation errors / warnings

olehermanse:
More fixes are needed to fix the actual issue (not just the log spamming), but this is a good improvement.

### DIFF
--- a/.github/workflows/add_validate_build_all.yaml
+++ b/.github/workflows/add_validate_build_all.yaml
@@ -1,0 +1,32 @@
+name: Add, validate, and build all modules
+
+# Controls when the workflow will run
+on:
+  # Triggers the workflow on master every night at 3:15 am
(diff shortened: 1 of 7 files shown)


.github/workflows/add_validate_build_all.yaml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
```
jakub-nt, line 25: the most recent version is v4

PR #256, opened 2025-07-15
Made validation more strict in cfbs add/remove/update/status/input

olehermanse:
This will probably warrant a major version bump since it seems relatively breaking, but I think we should do this, since it reduces the chances of someone editing a project into a state which their hub won't build correctly.


olehermanse:
Builds on https://github.com/cfengine/cfbs/pull/256

PR #265, opened 2025-08-04
Add the ability to specify custom build steps when adding modules

jakub-nt:


larsewi:
Please explain how this should work.

Do you need it for `cfbs convert`? If so, please explain why.

Is there a ticket? If so could you link it to the PR and in the last commit?

cfbs/cfbs_config.py
```diff
@@ -295,7 +296,7 @@ def _add_bundles_build_step(self, module, policy_files):
         module["steps"].append(step)
         log.debug("Added build step '%s' for module '%s'" % (step, name))
 
-    def _handle_local_module(self, module):
+    def _handle_local_module(self, module, use_default_build_steps=True):
```
olehermanse, line 299: I don't think it's necessary to switch from `explicit_build_steps` to `use_default_build_steps`. Just use `explicit_build_steps` all the way down where needed.

cfbs/cfbs_config.py
```diff
@@ -327,10 +328,12 @@ def _handle_local_module(self, module):
                     )
                 # TODO: Support adding local modules with autorun tag
 
-        self._add_policy_files_build_step(module)
-        self._add_bundles_build_step(module, policy_files)
+        if use_default_build_steps:
+            self._add_policy_files_build_step(module)
+            self._add_bundles_build_step(module, policy_files)
```
olehermanse, line 333: This leaves the module object in a "half done" state, I think one of these would be preferable:

1. Just add the defaults, and you can overwrite `module["steps"]` in the calling code.
2. Or, send `explicit_build_steps` all the way down to this function.

cfbs/cfbs_config.py
```diff
@@ -327,10 +328,12 @@ def _handle_local_module(self, module):
...
 
-    def _add_without_dependencies(self, modules):
+    def _add_without_dependencies(self, modules, use_default_build_steps=True):
+        """Note: `use_default_build_steps` is only relevant for local modules."""
```
olehermanse, line 336: This docstring note could be replaced by an assert.

cfbs/cfbs_config.py
```diff
@@ -362,7 +365,9 @@ def _add_modules(
         to_add: list,
         added_by="cfbs add",
         checksum=None,
+        explicit_build_steps: Optional[List[str]] = None,
     ):
+        """Note: explicit build steps are applied only to directly added modules, not their dependencies."""
```
olehermanse, line 370: Same here;

```python3
if explicit_build_steps is not None:
    assert len(to_add) == 1, "explicit_build_steps is only for adding a single module"
    assert is_local(to_add[0]), "explicit_build_steps is only for adding a local module"
```

olehermanse (changes requested):
Looks good, just some smaller things to fix.

cfbs/cfbs_config.py
jakub-nt: I would say the note is for the *users* of the function, for them to learn what the function parameter does. This is why it's in the docstring. An assert is only readable inside of the function.

olehermanse: You could have both. An assert is much more important IMO, and less likely to go stale.

cfbs/cfbs_config.py
jakub-nt: Yes, those are the methods. I still don't see what you would ideally like to see here, so feel free to suggest.

cfbs/cfbs_config.py
jakub-nt: The actual build step values would be unused then, there would only be a check whether they're `None` or not. That would read weird. Dunno.

PR #268, opened 2025-08-08
CFE-4476: cfbs-convert i1

jakub-nt:
Depends on #267 at the moment.


cf-bottom:
Thanks for submitting a PR! Maybe @craigcomstock can review this?

larsewi (approved):
Looks good 🚀 Would be nice if you could provide an example of you running the current iteration in the PR description or comment.

PR #281, opened 2025-09-19
CFE-4135: Rework and expand `cfbs show` output

jakub-nt:


larsewi (commented):
Looks good. Please fix it so that the Python 5 test passes

jakub-nt:
Note: The test failure is independent of this PR.

PR #304, opened 2026-02-17
ENT-12619: Reduced noise from cfbs shell test output

larsewi:
Example output:
```
$ bash tests/shell/all.sh
Warning: These shell based tests use the cfbs you have installed
         If you haven't already, run: pip install .
--- PASS: 001_init (2s)
--- PASS: 002_add (2s)
--- PASS: 003_download (2s)
--- PASS: 004_build (12s)
--- PASS: 005_alias (3s)
--- PASS: 006_search (2s)
--- SKIP: 007_install (0s)
--- PASS: 008_remove (2s)
 ---snip---
--- PASS: 047_absolute_path_modules (4s)

==============================
Test Results: 46 passed, 0 failed, 1 skipped (total: 47, 156s)
==============================
All cfbs shell tests completed successfully!
```

### DIFF
--- a/tests/shell/007_install.sh
+++ b/tests/shell/007_install.sh
@@ -1,10 +1,5 @@
-if [ x"$UNSAFE_TESTS" = x1 ]
-then
-    echo "Unsafe tests are enabled - running install test"
-else
-    echo "Warning: Unsafe tests are disabled - skipping install test"
-    exit 0
-fi
+source "$(dirname "$0")/common.sh"
+skip-unless-unsafe
 
 set -e
 set -x
--- a/tests/shell/all.sh
+++ b/tests/shell/all.sh
@@ -1,55 +1,108 @@
+#!/usr/bin/env bash
 echo "Warning: These shell based tests use the cfbs you have installed"
 echo "         If you haven't already, run: pip install ."
 set -e
-set -x
 export CFBS_USER_AGENT=CI # this user agent will be excluded from the build modules statistics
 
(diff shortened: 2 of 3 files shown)


craigcomstock (changes requested):
I can't say I love this PR. There is a lot of change and now the scripts can't be written free-hand in bash... you must refer to the testlib often to know what functions to use.

Also, I am not sure how much is improved other than maybe silencing "good" outputs to make analysis easier... which is a welcome change so +1 on that account.

I think all the migrated commits could be squashed into one for easier review... #next-time

larsewi:
@craigcomstock Thanks for the review :rocket: 

> I can't say I love this PR. There is a lot of change and now the scripts can't be written free-hand in bash... you must refer to the testlib often to know what functions to use.

I see your point, but I disagree to some extent. The scripts can still be written in free-hand bash. The asserts however are easier to understand for someone who is not very proficient with bash, rather than greps and pipes on top of each other. Furthermore, the asserts print out a nice error messages like `expected foo, got bar`, rather than test failing because grep returned 1.

> I think all the migrated commits could be squashed into one for easier review... #next-time

For me, personally, it's easier with smaller commits, rather than 600 lines of diff that I have to scroll through. You can always click the files changed tab to see everything together.

I will fix up some things. If you still think this PR is bad, I won't die on my sword.

craigcomstock (approved):
Pretty small change with some nice benefits.

PR #306, opened 2026-03-18
CFE-3948: Changed remove to either remove all dependencies, else nothing

SimonThalvorsen:
Ticket: CFE-3948

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -50,7 +50,6 @@ import re
 import copy
 import logging as log
 import json
-import functools
 from typing import Callable, List, Optional, Union
 from collections import OrderedDict
 from cfbs.analyze import analyze_policyset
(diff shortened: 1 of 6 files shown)


cfbs/commands.py
```diff
@@ -416,89 +415,141 @@ def remove_command(to_remove: List[str]):
...
+                d for d in _get_dependents(module_name) if d in modules_to_remove
+            ]
+            if not dependents_in_set:
+                module = _get_module_by_name(module_name)
+                if module:
+                    print("Removing module '%s'" % module_name)
+                    modules.remove(module)
+                    removed_modules.append(module)
+                    msg += "\n - Removed module '%s'" % module_name
+                modules_to_remove.remove(module_name)
```
larsewi: Modifying the set while iterating is generally not a good idea. Maybe it works accidentally because of the break afterwards.

cfbs/commands.py
```diff
@@ -416,89 +415,141 @@ def remove_command(to_remove: List[str]):
...
+                modules_to_remove.add(module["name"])
+                modules_to_remove.update(all_affected)
+            else:
+                print("Skipping removal of '%s'" % module["name"])
 
+    removed_modules = []
+    msg = ""
+    files = []
+
+    while modules_to_remove:
```
larsewi, line 515: Make sure this does not loop forever if the dependency graph has a cycle (A depends on B, B depends on A).

cfbs/commands.py
```diff
@@ -416,89 +415,141 @@ def remove_command(to_remove: List[str]):
...
+                        "Do you wish to continue?"
+                    ) % (
+                        module["name"],
+                        "s" if len(all_affected) > 1 else "",
+                        ", ".join(sorted(all_affected)),
+                        module["name"],
+                        "these modules" if len(all_affected) > 1 else "this module",
+                    )
+                else:
+                    prompt_msg = "Do you wish to remove '%s'?" % module["name"]
```
larsewi: This code is nearly identical below. Consider creating a helper function 

```Python3
def _prompt_for_removal(module_name, all_affected):
    pass
```

larsewi (changes requested):
Looking good 🚀 Just a few nitpicks

larsewi (approved):
Looks great 🚀

larsewi:
Maybe you could add a test?

tests/shell/049_remove_with_circular_dependencies/example-cfbs.json
```diff
@@ -0,0 +1,33 @@
...
+  "type": "policy-set",
+  "build": [
+    {
+      "name": "masterfiles-example-module",
+      "description": "Example",
+      "url": "https://github.com/example/",
+      "commit": "0000000000000000000000000000000000000000",
+      "added_by": "cfbs add",
+      "steps": ["run this"]
+    },
```
larsewi, line 13: Does this one need to be here for the test to work?

tests/shell/049_remove_with_circular_dependencies/example-cfbs.json
```diff
@@ -0,0 +1,33 @@
...
+    {
+      "name": "example-module",
+      "description": "Example",
+      "url": "https://github.com/example/",
+      "commit": "0000000000000000000000000000000000000000",
+      "dependencies": ["example-dependency"],
+      "added_by": "cfbs add",
+      "steps": ["delete this"]
```
larsewi, line 21: Are steps required in the validation, or could this be empty to reduce some of the noise?

tests/shell/049_remove_with_circular_dependencies.sh
```diff
@@ -0,0 +1,33 @@
...
+mkdir -p ./tmp/
+cd ./tmp/
+touch cfbs.json && rm cfbs.json
+rm -rf .git
+
+cp ../shell/049_remove_with_circular_dependencies/example-cfbs.json cfbs.json
+cfbs validate
+
+grep '"name": "example-module"' cfbs.json
+grep '"name": "example-dependency"' cfbs.json
```
larsewi, line 13: In this case module and dependency does not make sense since their both modules and dependencies. Maybe module-a and module-b would make more sense?

tests/shell/049_remove_with_circular_dependencies/example-cfbs.json
SimonThalvorsen: Required for validation

tests/shell/049_remove_with_circular_dependencies/example-cfbs.json
SimonThalvorsen: Yep, empty build does not validate correctly.

larsewi (approved):

PR #313, opened 2026-06-03
ENT-14170: Fixed updating to a pinned module version

nickanderson:
Previously doing something like cfbs update masterfiles@3.24.4 when running 3.24.1 would simply update to the latest version available in the index (while telling you it updated to the version you requested)

```
cfbs update masterfiles@3.24.4
    [main 8ab9378] Updated module 'masterfiles' from version 3.24.1 to version 3.24.4
     1 file changed, 3 insertions(+), 3 deletions(-)
    Updated module 'masterfiles' from version 3.24.1 to version 3.24.4?
jq -r '.build[] | select(.name=="masterfiles") | {version, commit}' cfbs.json
  {
    "version": "3.27.1",
    "commit": "15d8ca2ca951db3ddf197b90c9dea6ef75f5fa89"
  }
```

Ticket: ENT-14170

### DIFF
--- a/cfbs/index.py
+++ b/cfbs/index.py
@@ -245,6 +245,11 @@ class Index:
                 }
                 object.update(specifics)
                 object["version"] = version
+                # The version index stores an empty "subdirectory" for
+                # modules without one; an empty subdirectory is invalid, so
+                # drop it (matches how 'cfbs add' cleans the module up).
+                if object.get("subdirectory") == "":
+                    del object["subdirectory"]
 
         assert object is not None
         module.update(object)
--- a/tests/shell/050_update_masterfiles_specific_version.sh
+++ b/tests/shell/050_update_masterfiles_specific_version.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+set -e
+set -x
+cd tests/
+mkdir -p ./tmp/
+cd ./tmp/
+touch cfbs.json && rm cfbs.json
+rm -rf .git
+
+# Inspect the masterfiles module specifically, rather than grepping the whole
+# file (which could match a field belonging to a different module).
+mf() { jq -r --arg k "$1" '.build[] | select(.name == "masterfiles") | .[$k]' cfbs.json; }
+mf_has() { jq -r --arg k "$1" '.build[] | select(.name == "masterfiles") | has($k)' cfbs.json; }
+
+# Start on an old version of masterfiles.
+cfbs --non-interactive init --masterfiles=3.24.1
+[ "$(mf version)" = "3.24.1" ]
+[ "$(mf commit)" = "1171e2e50a229d78e2fdd4357a5d07ecc19bdbf4" ]
+
+# Update to a specific intermediate version (not the latest). This must pin to
+# exactly 3.24.4, not jump to the newest release in the index.
+cfbs --non-interactive update masterfiles@3.24.4
+[ "$(mf version)" = "3.24.4" ]
+[ "$(mf commit)" = "ed4628805e352fd68d7d72664c859df5a4bb0715" ]
+# Why we assert on "subdirectory" here: pinning an update to a version now routes
+# through index.get_module_object() *with that version*, which reads the
+# per-version index. That index stores an empty "subdirectory" ("") for modules
+# that don't have one, and an empty subdirectory is invalid and breaks validation
+# and build. So get_module_object() now has to drop it (matching how 'cfbs add'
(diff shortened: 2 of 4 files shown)


nickanderson:
First commit failed the test as expected, second commit made it pass.

tests/shell/050_update_masterfiles_specific_version.sh
```diff
@@ -0,0 +1,42 @@
...
+[ "$(mf_has subdirectory)" = "false" ]
+cfbs validate
+
+# Asking again for the same version is a no-op.
+cfbs --non-interactive update masterfiles@3.24.4
+[ "$(mf version)" = "3.24.4" ]
+
+# Asking for an older version than the current one must not downgrade.
+cfbs --non-interactive update masterfiles@3.24.1
+[ "$(mf version)" = "3.24.4" ]
```
nickanderson, line 41: I am unsure that this is exactly desired behavior ^

tests/shell/050_update_masterfiles_specific_version.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# A plain update (no @version) still moves to a newer version than the pin.
+cfbs --non-interactive update masterfiles
+[ "$(mf version)" != "3.24.4" ]
```
craigcomstock: This certainly doesn't check for "newer" only "not" so it would seem you might want to check for newer actually.

tests/shell/050_update_masterfiles_specific_version.sh
craigcomstock: @olehermanse why does "update" allow a version at all? It would seem to me that the update subcommand might just fail if a version with `@version` is specified.

If that is the case then I suppose `cfbs add` would have to be used to change the version of an already installed module?

tests/shell/050_update_masterfiles_specific_version.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+set -e
+set -x
+cd tests/
+mkdir -p ./tmp/
```
craigcomstock, line 5: I know this is cargo cult here and "already works" but I am wondering how these dirs are managed from shell test to shell test. Seems like you would be making the same tmp dir over and over and have junk in them from test to test. I suppose it doesn't matter because we rm cfbs.json and .git and re-make a project so I imagine random files could be laying around if we create them but they won't be included in the cfbs build output. OK.

tests/shell/050_update_masterfiles_specific_version.sh
```diff
@@ -0,0 +1,42 @@
...
+[ "$(mf commit)" = "1171e2e50a229d78e2fdd4357a5d07ecc19bdbf4" ]
+
+# Update to a specific intermediate version (not the latest). This must pin to
+# exactly 3.24.4, not jump to the newest release in the index.
+cfbs --non-interactive update masterfiles@3.24.4
+[ "$(mf version)" = "3.24.4" ]
+[ "$(mf commit)" = "ed4628805e352fd68d7d72664c859df5a4bb0715" ]
+# The empty "subdirectory" from the version index must not leak in (it would be
+# invalid and break validation / build).
+[ "$(mf_has subdirectory)" = "false" ]
```
craigcomstock, line 32: why is this here? doesn't seem related to this test case.

cfbs/index.py
```diff
@@ -245,6 +245,11 @@ def get_module_object(
                 }
                 object.update(specifics)
                 object["version"] = version
+                # The version index stores an empty "subdirectory" for
+                # modules without one; an empty subdirectory is invalid, so
+                # drop it (matches how 'cfbs add' cleans the module up).
+                if object.get("subdirectory") == "":
```
craigcomstock, line 251: I see now why you have this in your test. You likely added the first fix in commands.py and found that the build failed. Ok.

craigcomstock: You could add a comment to the test saying WHY you are testing that... because when doing "update module" now... the behavior of get_module_object() had to change.

craigcomstock: So maybe we should make a ticket to remove the empty subdirectory and fix the "source" to be correct so we can avoid hacks like this one.

craigcomstock (approved):

tests/shell/050_update_masterfiles_specific_version.sh
nickanderson: splained

craigcomstock (approved):
squash the commits yeah?

PR #314, opened 2026-06-04
ENT-14174: Changed cfbs update module@version to allow downgrades

nickanderson:
Given the target for reproducible builds, it doesn't seem quite reasonable to
disallow downgrades. Allowing this actually facilitates demonstrations of the
tool in use.

Ticket: ENT-14174
Changelog: Title

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -740,11 +740,9 @@ def update_command(to_update):
                 continue
             elif local_ver > index_ver:
                 log.warning(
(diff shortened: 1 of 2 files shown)


craigcomstock (approved):
yeah!

olehermanse:
> Given the target for reproducible builds, it doesn't seem quite reasonable to
disallow downgrades. Allowing this actually facilitates demonstrations of the
tool in use.

I don't really understand the logic here :smile:

olehermanse (approved):

PR #324, opened 2026-07-22
ENT-14302: Added a check for .git-dir inside masterfiles-* when using `cfbs convert` to avoid crashing

SimonThalvorsen:
Ticket: ENT-14302

### DIFF
--- a/cfbs/commands.py
+++ b/cfbs/commands.py
@@ -1225,6 +1225,15 @@ def convert_command(non_interactive=False, offline=False):
 
     dir_name = dir_content[0]
     path_string = "./" + dir_name + "/"
+    has_gitfolder = os.path.isdir(os.path.join(path_string, ".git"))
+    if has_gitfolder:
+        print(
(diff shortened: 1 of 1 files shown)


larsewi (commented):
Should it perhaps just skip initializing a new git repository instead of asking you to remove the existing `.git` directory? I'm guessing that if you want to convert an existing git repository, you'd still want to keep the original git history.

olehermanse (approved):

olehermanse:
> Should it perhaps just skip initializing a new git repository instead of asking you to remove the existing .git directory? I'm guessing that if you want to convert an existing git repository, you'd still want to keep the original git history.

That would require a bit more work - need to "rename" all the existing files in git, make another commit, so that they are in a subdir, etc.

Also, depends on what the user wants. Some users might want to "start fresh" in this case, while others will want to convert / retain their original git repo.

https://northerntech.atlassian.net/browse/CFE-4700

PR #326, opened 2026-07-28
ENT-11346: Added the 'cfbs render-input' command

larsewi:
Converts input data for a module into an augments file (`def.json`). It does the same conversion as the `input` build step performs during `cfbs build`, except that the input data is read from an infile and nothing is stored in the built project.

Mission Portal needs this to render the augment for module input which is entered per host group, and thus stored in its database rather than in the project.

```
$ cfbs init
 ---snip---
$ cfbs get-input uninstall-packages - | jq '.[0].response += [{"name": "curl", "why": "its cool"}]' | cfbs render-input uninstall-packages - -
{
  "variables": {
    "uninstall_packages:uninstall_packages.package_names": {
      "value": [{ "name": "curl", "why": "its cool" }],
      "comment": "Added by 'cfbs input'"
    }
  }
}
```

The first 3 commits are preparatory refactoring, with no functional change. The next 3 fix pre-existing crashes found while implementing this, each with a shell test which fails without its fix.

Ticket: ENT-11346

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cfbs/main.py
+++ b/cfbs/main.py
@@ -240,6 +240,42 @@ def _main() -> int:
         finally:
             if needs_close:
                 file.close()
+    if args.command == "render-input":
+        if len(args.args) != 3:
+            log.error(
+                "%s <module>, <infile (or - for stdin)>"
+                " and <outfile (or - for stdout)>"
+                % (
+                    "Too many arguments: expected"
+                    if len(args.args) > 3
+                    else "Missing required arguments"
+                )
+            )
+            return 1
+
+        module, infilename, outfilename = args.args
+
+        # Open the infile first, so that a mistyped infile doesn't truncate the
+        # outfile:
+        try:
+            infile, close_infile = open_file_arg(infilename, "r")
+        except OSError as e:
+            log.error("Can't open '%s': %s" % (infilename, e))
+            return 1
+        try:
+            outfile, close_outfile = open_file_arg(outfilename, "w")
+        except OSError as e:
+            log.error("Can't open '%s': %s" % (outfilename, e))
+            if close_infile:
+                infile.close()
+            return 1
+        try:
+            return commands.render_input_command(module, infile, outfile)
+        finally:
+            if close_infile:
+                infile.close()
+            if close_outfile:
+                outfile.close()
 
     raise CFBSProgrammerError(
         "Command '%s' not handled appropriately by the code above" % args.command
--- a/cfbs/main.py
+++ b/cfbs/main.py
@@ -4,7 +4,6 @@
 __copyright__ = ["Northern.tech AS"]
 
 import logging as log
-import sys
 import os
 import traceback
 import pathlib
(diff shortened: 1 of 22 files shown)


tests/shell/051_get_input_no_build_list.sh
```diff
@@ -0,0 +1,54 @@
+set -e
+set -x
+cd tests/
+mkdir -p ./tmp/
+cd ./tmp/
+touch cfbs.json && rm cfbs.json
+rm -rf .git
+rm -rf delete-files
+
+# A cfbs.json for a module, which uses "provides" and thus has no "build"
```
olehermanse: `provides` does not imply or mandate that there is no `build`, so the `thus` here is wrong and misleading.

cfbs/main.py
```diff
@@ -240,6 +240,42 @@ def _main() -> int:
...
+            if close_infile:
+                infile.close()
+            return 1
+        try:
+            return commands.render_input_command(module, infile, outfile)
+        finally:
+            if close_infile:
+                infile.close()
+            if close_outfile:
+                outfile.close()
```
olehermanse: Manually closing files is easy to forget or mess up, why not use `with` context managers?

olehermanse (approved):

cfbs/main.py
```diff
@@ -226,20 +227,50 @@ def _main() -> int:
...
+                    if len(args.args) > 3
+                    else "Missing required arguments"
+                )
+            )
+            return 1
+
+        module, infilename, outfilename = args.args
+
+        # ExitStack rather than a plain with statement, so that the OSError
+        # handling only covers opening the file, not running the command:
```
olehermanse:
```suggestion
        # handling only covers opening the files, not whatever exceptions happen in commands.render_input_command():
```

cfbs/main.py
```diff
@@ -226,20 +227,50 @@ def _main() -> int:
...
+                infile = stack.enter_context(open_file_arg(infilename, "r"))
+            except OSError as e:
+                log.error("Can't open '%s': %s" % (infilename, e))
+                return 1
+            try:
+                outfile = stack.enter_context(open_file_arg(outfilename, "w"))
+            except OSError as e:
+                log.error("Can't open '%s': %s" % (outfilename, e))
+                return 1
+            return commands.render_input_command(module, infile, outfile)
```
olehermanse, line 275: Looks a bit unnecessarily complicated IMO, don't know exactly what's the best way.

Something like this feels more natural for me;


```suggestion
        unexpected = None
        
        try:
            with open(inp, "r") as infile:
                with open(out, "w") as outfile:
                    try:
                        commands.render_input_command(module, infile, outfile)
                    except Exception as e:
                        unexpected = e
        except OSError as e:
            log.error("Can't open '%s': %s" % (outfilename, e))
            return 1
        
        if unexpected:
            raise unexpected
```

olehermanse (approved):

cfbs/main.py
larsewi: Yeah, I think both versions looks complicated

PR #328, opened 2026-07-31
ENT-14368: Added a new input-type: string-multiline

SimonThalvorsen:
Ticket: ENT-14368

### DIFF
--- a/cfbs/cfbs_config.py
+++ b/cfbs/cfbs_config.py
@@ -44,7 +44,7 @@ from cfbs.module import (
     is_module_local,
     is_module_absolute,
 )
-from cfbs.prompts import prompt_user, prompt_user_yesno
+from cfbs.prompts import prompt_user, prompt_user_yesno, prompt_user_multiline
 from cfbs.validate import validate_single_module
 
 
@@ -565,6 +565,15 @@ class CFBSConfig(CFBSJson):
             )
             return response
 
+        def _input_multiline_string(input_data):
+            _check_keys(["question"], input_data)
+            response = prompt_user_multiline(
+                self.non_interactive,
+                input_data["question"],
+                default=input_data.get("default"),
+            )
+            return response
+
         def _input_elements(subtype):
             result = OrderedDict()
             for element in subtype:
@@ -615,6 +624,8 @@ class CFBSConfig(CFBSJson):
 
             if definition["type"] == "string":
                 definition["response"] = _input_string(definition)
+            elif definition["type"] == "string-multiline":
+                definition["response"] = _input_multiline_string(definition)
             elif definition["type"] == "list":
                 definition["response"] = _input_list(definition)
             else:
--- a/cfbs/prompts.py
+++ b/cfbs/prompts.py
@@ -43,6 +43,40 @@ def prompt_user(non_interactive: bool, prompt: str, choices=None, default=None):
     return answer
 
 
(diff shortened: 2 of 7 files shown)


tests/shell/059_input_multilinestring/example-cfbs.json
```diff
@@ -0,0 +1,33 @@
...
+          "type": "string",
+          "variable": "filename",
+          "namespace": "cfbs",
+          "bundle": "create_single_file_with_content",
+          "label": "Filename",
+          "question": "What file should this module create?",
+          "default": "/tmp/create-single-file-with-content.txt"
+        },
+        {
+          "type": "multilinestring",
```
olehermanse: Let's maybe do this to be more prepared for other type variants in the future;


```suggestion
          "type": "string-multiline",
```

cfbs/cfbs_config.py
```diff
@@ -615,6 +624,8 @@ def _input_list(input_data):
 
             if definition["type"] == "string":
                 definition["response"] = _input_string(definition)
+            elif definition["type"] == "multilinestring":
```
olehermanse:
```suggestion
            elif definition["type"] == "string-multiline":
```

cfbs/validate.py
```diff
@@ -682,10 +682,10 @@ def _validate_module_input(name, module):
                     % field,
                 )
 
-        if input_element["type"] not in ("string", "list"):
+        if input_element["type"] not in ("string", "list", "multilinestring"):
```
olehermanse:
```suggestion
        if input_element["type"] not in ("string", "list", "string-multiline"):
```

cfbs/validate.py
```diff
@@ -682,10 +682,10 @@ def _validate_module_input(name, module):
...
             raise CFBSValidationError(
                 name,
-                'The input "type" must be "string" or "list", not "%s"'
+                'The input "type" must be "string", "multilinestring", or "list", not "%s"'
```
olehermanse:
```suggestion
                'The input "type" must be "string", "string-multiline", or "list", not "%s"'
```

olehermanse (changes requested):

tests/shell/059_input_multilinestring/example-cfbs.json
olehermanse: We already use dashes in project type, so that seems nice to follow as a convention.

cfbs/prompts.py
```diff
@@ -43,6 +43,40 @@ def prompt_user(non_interactive: bool, prompt: str, choices=None, default=None):
...
+            raise ValueError(
+                "Missing default value for prompt '%s' in non-interactive mode" % prompt
+            )
+        return default
+
+    print(prompt)
+    print(
+        "(Enter one or more lines of text, then finish with double newline or Ctrl+D"
+        " (Ctrl+Z followed by Enter on Windows))"
+    )
```
larsewi, line 58: What if you try to detect if you are on Windows and print a different version of this hint based on the platform?

olehermanse (approved):

larsewi (commented):
Maybe
```
mv tests/shell/059_input_multilinestring.sh tests/shell/059_input_string_multiline.sh
mv tests/shell/059_input_multilinestring/ tests/shell/059_input_string_multiline/
```
?

PR #330, opened 2026-08-03
ENT-4074: Added a new input-type: file

SimonThalvorsen:
Ticket: ENT-4074


cf-bottom:
Thank you for submitting a pull request! Maybe @craigcomstock can review this?

cfbs/build.py
```diff
@@ -243,7 +243,81 @@ def _perform_directory_step(args, source, destination, prefix):
...
+            operation, args = split_build_step(step)
+            if operation != "directory" or len(args) != 2:
+                continue
+            src, dst = args
+            if src not in (".", "./"):
+                continue
+            rel = os.path.relpath(abs_path, module_root)
+            dst = "" if dst in (".", "./") else dst
+            dest = os.path.join(destination, dst, rel)
+            return "$(sys.policy_entry_dirname)/" + os.path.relpath(dest, destination)
```
larsewi: `$(sys.policy_entry_dirname)` would resolve to `/var/cfengine/inputs/` on a normal agent run. Because it would run `promises.cf` which is in that directory. However, if you e.g. execute a policy file containing the `main` bundle and execute it from your home directory, then this path would resolve to you home directory.

Maybe the correct variable to use here would be `$(sys.input_dir)/` ?

cfbs/build.py
```diff
@@ -243,7 +243,81 @@ def _perform_directory_step(args, source, destination, prefix):
     write_json(defjson, merged)
 
 
-def _perform_input_step(args, name, destination, prefix):
+def _already_shipped_destination(path, build_modules, destination):
```
larsewi: Function name and docs string is confusing. It's not obvious that this function returns a path based on the function name.

cfbs/build.py
```diff
@@ -243,7 +243,81 @@ def _perform_directory_step(args, source, destination, prefix):
...
+    copy the file itself - just point at where it will end up.
+    """
+    abs_path = os.path.abspath(path)
+    for module in build_modules:
+        module_name = module.get("name", "")
+        if not (module_name.startswith("./") and module_name.endswith("/")):
+            continue
+        module_root = os.path.abspath(module_name)
+        if os.path.commonpath([abs_path, module_root]) != module_root:
+            continue
```
larsewi, line 265: A comment here would be nice

cfbs/build.py
```diff
@@ -243,7 +243,81 @@ def _perform_directory_step(args, source, destination, prefix):
     write_json(defjson, merged)
 
 
-def _perform_input_step(args, name, destination, prefix):
+def _path_if_already_shipped(path, build_modules, destination):
```
larsewi, line 246: So this function both tests if the file is already shipped and computes the if not? It feels a bit messy, but I don't know how to make it better.

JSON.md
```diff
@@ -855,3 +855,120 @@ $ cat ./out/masterfiles/def.json
...
+]
+```
+
+Without `"while"`, `"response"` is a single path instead of a list.
+A file already inside the project is referred to as-is; a file from outside (as above) is copied into the module's own directory, next to `input.json`.
+
+
+Sometimes you may want to place all your files in a separate directory for easier management.
+```
+$ eza -T --level 2
```
larsewi: I don't think most people are familiar with this command. Still fine to render the output. Or you could maybe use `tree`?
```suggestion
```

JSON.md
```diff
@@ -855,3 +855,120 @@ $ cat ./out/masterfiles/def.json
...
+--snip--
+
+$ cat out/masterfiles/def.json
+{
+  "inputs": ["services/cfbs/modules/run-script-module/main.cf"],
+  "vars": { "control_common_bundlesequence_end": ["run_script_module:main"] },
+  "variables": {
+    "run_script_module:main.scripts": {
+      "value": [
+        "$(sys.inputdir)/services/cfbs/scripts/script1.sh",
```
larsewi, line 952: So the file input paths end up in the augments? Should you be able to add files without having them modifying augments? E.g., by omitting `variable`, `namespace`, `bundle`?

tests/shell/060_input_file.sh
```diff
@@ -0,0 +1,48 @@
+set -e
```
larsewi, line 1: Please add a test here testing no extension constraints.

JSON.md
SimonThalvorsen: I personally do not think so, my understanding of how it should work is that when using `cfbs input` the response should be available from the augment. If you would want to just transfer files, then you should just add a directory containing them as a local module

JSON.md
```diff
@@ -855,3 +855,126 @@ $ cat ./out/masterfiles/def.json
...
+    "while": "Do you want to add another script?",
+    "response": ["./run-scripts/deploy.sh", "./run-scripts/rollback.sh"]
+  }
+]
+```
+
+Without `"while"`, `"response"` is a single path instead of a list.
+A file already inside the project is referred to as-is; a file from outside (as above) is copied into the module's own directory, next to `input.json`.
+
+
```
larsewi:
```suggestion

```

JSON.md
```diff
@@ -855,3 +855,126 @@ $ cat ./out/masterfiles/def.json
...
+Sometimes you may want to place all your files in a separate directory for easier management.
+```
+$ tree --level 2
```
larsewi:
```suggestion
$ tree -L 2
```

```
$ tree --level 2
tree: Invalid argument `--level'.
usage: tree [-acdfghilnpqrstuvxACDFJQNSUX] [-L level [-R]] [-H [-]baseHREF]
	[-T title] [-o filename] [-P pattern] [-I pattern] [--gitignore]
	[--gitfile[=]file] [--matchdirs] [--metafirst] [--ignore-case]
	[--nolinks] [--hintro[=]file] [--houtro[=]file] [--inodes] [--device]
	[--sort[=]name] [--dirsfirst] [--filesfirst] [--filelimit[=]#] [--si]
	[--du] [--prune] [--charset[=]X] [--timefmt[=]format] [--fromfile]
	[--fromtabfile] [--fflinks] [--info] [--infofile[=]file] [--noreport]
	[--hyperlink] [--scheme[=]schema] [--authority[=]host] [--opt-toggle]
	[--version] [--help] [--] [directory ...]
```

Please go through your example and make sure it works ;)

tests/shell/060_input_file.sh
```diff
@@ -0,0 +1,55 @@
...
+cfbs render-input copy-a-file copy-a-file/input.json actual.output
+diff actual.output ../shell/060_input_file/expected-augment.json
+
+# A "file" input with a "while" prompt must let the user supply multiple
+# files. Files from outside the project must be copied into the module's
+# directory, next to input.json, and "response" updated to a list of the
+# (possibly localized) paths:
+echo "echo one" > /tmp/one.sh
+echo "echo two" > /tmp/two.sh
+printf '/tmp/one.sh\nyes\n/tmp/two.sh\nno\n' | cfbs input run-scripts
```
larsewi, line 42: What happens if you pick `/tmp/a/one.sh` and `/tmp/b/one.sh`?

SimonThalvorsen: Then you would get two entries for one.sh ... both referring to the last added file

larsewi: So the last file would overwrite the first file? Furthermore, it can overwrite an already existing file in the module directory if they happened to have the same basename? This should probably error instead of overwriting existing files?

